PR #54, opened 2015-03-28
Take into account __future__ imports and pass flags to compile()

### LINKED ISSUE #53: Future imports aren't taken into account when checking syntax
This means that code like this will cause a SyntaxError:

	from __future__ import print_function

	def call_my_function(the_function):
	    the_function("hi")

	if __name__ == "__main__":
	    call_my_function(print)

with output/traceback like so:

	INTERNAL ERROR: 
	    call_my_function(print)
	  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
	    "__main__", fname, loader, pkg_name)
	  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
	    exec code in run_globals
	  File "/home/flyte/dev/yapf/yapf/__main__.py", line 18, in <module>
	    sys.exit(yapf.main(sys.argv))
	  File "yapf/__init__.py", line 102, in main
	    print_diff=args.diff)
	  File "yapf/__init__.py", line 124, in FormatFiles
	    filename, style_config=style_config, lines=lines, print_diff=print_diff)
	  File "yapf/yapflib/yapf_api.py", line 68, in FormatFile
	    print_diff=print_diff)
	  File "yapf/yapflib/yapf_api.py", line 113, in FormatCode
	    reformatted_source = reformatter.Reformat(uwlines)
	  File "yapf/yapflib/reformatter.py", line 73, in Reformat
	    verifier.VerifyCode(formatted_code[-1])
	  File "yapf/yapflib/verifier.py", line 45, in VerifyCode
	    compile(normalized_code.encode('UTF-8'), '<string>', 'exec')
	  File "<string>", line 1
	    call_my_function(print)
		                     ^
	SyntaxError: invalid syntax

The problem is that each line is taken on its own merit without taking into account imports from __future__.

flyte:
Fixes #53 by parsing the source for __future__ imports, then supplying the correct flags to compile().

This has the side-effect of changing the function signature for reformatter.Reformat() by allowing the unformatted_source to be passed in. This may not be ideal, so please let me know if you can see a more suitable way to get the compiler flags from where they're extracted from the source to the compile() calls in verifier.VerifyCode().

I have accepted the CLA.

### DIFF
--- a/yapf/yapflib/reformatter.py
+++ b/yapf/yapflib/reformatter.py
@@ -20,10 +20,16 @@ as a string.
   Reformat(): the main function exported by this module.
 """
 
+import __future__
+import tokenize
 import collections
 import heapq
 import re
 
+try:
+  from StringIO import StringIO
+except ImportError:
+  from io import StringIO
 from yapf.yapflib import format_decision_state
 from yapf.yapflib import line_joiner
 from yapf.yapflib import pytree_utils
@@ -31,7 +37,7 @@ from yapf.yapflib import style
 from yapf.yapflib import verifier
 
 
-def Reformat(uwlines):
+def Reformat(uwlines, unformatted_source=None):
   """Reformat the unwrapped lines.
 
   Arguments:
@@ -42,6 +48,8 @@ def Reformat(uwlines):
   """
   final_lines = []
   prev_last_uwline = None  # The previous line.
+  compile_flags = _ExtractFutureImports(unformatted_source) \
+                  if unformatted_source else 0
 
   for uwline in _SingleOrMergedLines(uwlines):
     first_token = uwline.first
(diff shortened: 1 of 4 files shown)


gwelymernans:
Hi Flyte,

Thanks for the code. I looked over it and I'm not sure that we need all of this for what we're doing. Or at the least it should be done in a different place. :-)

The idea of the "verifier" module is that it will catch catastrophic errors. Note that there's a better way to validate that the semantics haven't changed (generating before and after ASTs and comparing them). So with that in mind, this seems like a bit of overkill. An easier way to do this is to generate all of the supported "from __future__ import ..." statements and try to compile / parse them in the verifier module. We do something similar with the `print` statement already.

flyte:
Understood. I shouldn't have stormed ahead with it without discussing it here first anyway.. I'll probably turn the code into a library since I couldn't find any to do this. Thanks for your responses.

flyte:
Just if you're interested, I looked into the AST you mentioned and have used it to simplify my code massively. I did turn it into a library but it's hardly worth it now that it's so easy!
https://github.com/flyte/static-future/blob/develop/static_future/analysis.py

gwelymernans:
A side note, we're thinking about disabling the verification step for all but development (and maybe not even that).

PR #72, opened 2015-04-06
Improve handling of filenames on the command line

sbc100:
Don't call GetCommandLineFiles unless there are positional
arguments (args.files), and only pass the positional parameters
(otherwise GetCommandLineFiles ends up processing all the args
as well).

Prior to this change if you run 'yapf -i foo.c' it will give you
the unhelpful error:
'cannot use --in_place or --diff flags when reading from stdin'

Add a unit test for this case. Would add here but would prefer
to use 'mock' and don't see a dependency on 'mock' yet.

Also, give a useful error message when positional paramters
are specified but not arguments were given.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -11,3 +11,4 @@
 
 Bill Wendling <morbo@google.com>
 Eli Bendersky <eliben@google.com>
+Sam Clegg <sbc@google.com>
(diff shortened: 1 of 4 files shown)


sbc100:
My google and github accounts should be linked already.

gwelymernans:
Odd. Could you go to https://cla.developers.google.com/  and see if the CLA is listed? If you can't find it, you may have to resign. :-(

sbc100:
Its my fault.. i should have used my @google.com address.  Will upload a new version of PR shortly.

sbc100:
Should be good now.

PR #81, opened 2015-04-09
Move builtin styles into config files

sbc100:
This change is quite a bit more intrusive than my previous few.  Feel free to push back if you don't like it.   It certainly makes sense to me this way (i didn't expect the style definitions in code when I first poked around the project).  On thing I wasn't sure about was the file extension to use for styles.  I went with .ini because this caused vim to choose the right syntax highlighting :)

For my next change I was thinking of having yapf look for a project wide style file (.yapf) that would work like .clang-format does..  again not sure if that fits with your direction, but I thought it would be useful for yapf itself for example.

Move the builtin styles into config files rather than hardcoding
them.  This allows them to be easily located and modified and
also to act as an example to those who want to write their own
styles.

It also keeps us honest and ensures we don't encode anything into
the builtin styles that cannot be parsed from the config file format.

It also allows new builtin styles to be added by simply dropping
files in the 'styles' directory.

### DIFF
--- a/yapf/styles/pep8.ini
+++ b/yapf/styles/pep8.ini
@@ -1,0 +1,71 @@
+[style]
+
+# Align closing bracket with visual indentation.
+align_closing_bracket_with_visual_indent = true
+
+# The column limit.
+column_limit = 79
+
+# The regex for an i18n comment. The presence of this comment stops
+# reformatting of that line, because the comments are required to be
+# next to the string they translate.
+i18n_comment = ''
+
+# The i18n function call names. The presence of this function stops
+# reformattting on that line, because the string it has cannot be moved
+# away from the i18n comment.
+i18n_function_call = ''
+
+# The number of columns to use for indentation.
+indent_width = 4
+
+# Indent width for line continuations.
+continuation_indent_width = 4
+
+# Insert a blank line before a 'def' or 'class' immediately nested within
+# another 'def' or 'class'.
+#
+# For example:
+#
+# class Foo:
+#                    # <------ this blank line
+#   def method():
+#     ...
+#
+blank_line_before_nested_class_or_def = false
+
+# Insert a space between the ending comma and closing bracket of a list
+# etc.
+space_between_ending_comma_and_closing_bracket = true
+
+# The number of spaces required before a trailing comment.
+spaces_before_comment = 2
+
+# Set to true to prefer splitting before 'and' or 'or' rather than
+# after.
+split_before_logical_operator = false
+
+# Split named assignments onto individual lines.
+split_before_named_assigns = true
+
+# The penalty for splitting the line after a unary operator.
+split_penalty_after_unary_operator = 100
+
+# The penalty for characters over the column limit.
+split_penalty_excess_character = 2000
+
+# The penalty of splitting the line around the 'and' and 'or' operators.
+split_penalty_logical_operator = 30
+
+# The penalty for not matching the splitting decision for the matching
+# bracket tokens. For instance, if there is a newline after the opening
+# bracket, we would tend to expect one before the closing bracket, and
+# vice versa.
+split_penalty_matching_bracket = 50
+
+# The penalty for splitting right after the opening bracket.
+split_penalty_after_opening_bracket = 30
+
+# The penalty incurred by adding a line split to the unwrapped line. The
+# more line splits added the higher the penalty.
+split_penalty_for_added_line_split = 30
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,7 @@ with codecs.open('README.rst', 'r', 'utf-8') as fd:
       author='Google Inc.',
       maintainer='Bill Wendling',
       maintainer_email='morbo@google.com',
+      package_data={'yapf': ['styles/*.ini']},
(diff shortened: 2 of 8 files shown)


eliben:
On the flip side, this creates a distribution issue, wherein yapf has to find these .ini files when it runs in all scenarios - installed, as a script, as a library, and so on. I'd prefer to avoid this, to be honest.

To provide an example of how to create new styles it's possible to just create a test with that, or some sort of sample. 

FWIW, I think `clang-format` also hard-codes its built-in styles, for the same reason. It's very nice to be able to just run the `clang-format` executable without worrying about how it's going to find its data files. Data files are usually a pain in the neck.

sbc100:
I know that some platforms have distribution issues when shipping data files, but I don't see any here.  What specific scenarios do you envisage causing problems?  This method (or finding the styles folder relative the .py file itself) works fine when you run the yapf in-place, or install if via setup.py.   Maybe I'm missing something?

sbc100:
Indeed, isn't that the entire purpose of the 'package_data' field in setup.py?

eliben:
I don't have a concrete problem to point out. Just a bunch of bad experience with distributing data files, especially *when your code is used as a library*, which is one of the intended use cases of yapf.

When some other project uses your library, you create distribution issues for it by requiring to carry data files around too.

I just don't see much benefit in doing this, to be honest. I'll let @gwelymernans decide.

eliben:
Sam, take a look at #100 which suggests an alternative to your .yapf file proposal

hayd:
I like this proposal, though IMO `.yapf` or `.yapfstyle` is better than `.ini`.

These make great templates and documentation how yapf config files (as in #100) should work, it seems very elegant. :)

hayd:
or perhaps style.yapf (for a local config, in your package's root directory) and pep8.yapf, google.yapf, etc.

sbc100:
I like $PROJECT_ROOT/.style.yapf.  I'll make that a separate change.

I terms of this change, Eli's argument is that it could make the library harder to package/use, and that makes some amount of sense to me.  Right now I rely in __file__ to find the data files relative the library.  This works for all the use cases I have tried:

- point PYTHONPATH directly and source checkout
- install via ./setup.py
- pip install
- cp -ar yapf $MY_PROJECT

I can imagine some other use cases that I have not tried.  In particular running/install in other runtimes such as pypy, or one of the .net python implementations.  However, the only was I can see this being a problem is if the implementation doesn't correctly support __file__, which would be very broken.  I guess they might have their own packaging system and they could forget to package the data files, but I don't that is a big deal.

I'll make your change to use a different file extension, and I'll make sure that if the data files are somehow not found that a meaningful error is produced.

hayd:
It looks like if you use `__file__`  you may have to set `zip_safe = False` in setup.py ?

(A potential workaround which I don't understand: https://github.com/joke2k/faker/pull/212/files)

sbc100:
Oh.. I've not seen the zip_safe thing .. that could be an argument against my change.  I will investigate.

sbc100:
Re-based onto master

yapf/styles/pep8.ini
```diff
@@ -0,0 +1,71 @@
+[style]
+
+# Align closing bracket with visual indentation.
+align_closing_bracket_with_visual_indent = true
+
+# The column limit.
```
hayd: These comments are now duplicated with the help in style.py (so they could go out of sync, not really a problem, but I think they could be removed).

gwelymernans:
For what it's worth, I prefer the direction that #108 is going rather than having `.ini` files within the codebase / distribution. If I remember correctly, `clang-format` doesn't use `.ini` files either.

hayd:
I don't think yapf is currently `zip_safe` since the (testing) code already includes a `__file__`.

@gwelymernans Are you against all data files, or would .yapf over .ini be better?

gwelymernans:
What kind of `.yapf` file? I think that #108 was going to add a style file per project or something like that.

hayd:
@gwelymernans .yapf would be an .ini in all but name (as would the .style.yapf of #108 ).

This PR implements pep8, chromium and google styles as if they were .style.yapf files i.e. these default styles are implemented via .yapf files.

IMO the main argument for this is "it's elegant", maybe that isn't so compelling... :(

sbc100:
My plan is to revisit this once #108 lands.

eliben:
@sbc100 before you invest time in revisiting this, what is the plan? 

I think both @gwelymernans and I are unhappy with making yapf depend on data files without there being a really good reason for it.  It seems that the currently stated reason is mostly aesthetical, which doesn't pass the bar.

I'll close this PR soon unless there's new information.

jamesbroadhead:
as a new user: I strongly support this PR as a way to document the config format & make it harder to special-case the built-in styles.

Daenyth:
As a potential alternative, what about keeping the defaults in code but adding a `-dump-config` option like clang-format has?

gwelymernans:
I have no problem with a `--dump-config` option. Would you like to do the honors?

Daenyth:
I might do it if I have time, but honestly it's not at the top of my list.

PR #93, opened 2015-04-13
Add --max-line-width and --indent-width

alexjurkiewicz:
I know `--style` supports dictionary overrides, but it's pretty unintuitive. Adding line-width as a first-class commandline option matches pep8 the linter. Indent width was easy to add so I did that too, but can remove if you prefer.

PS, these options have a few synonyms. "line width", "column width", "line length" are all in use, as is "indent size" and "indent width". I picked names that seemed to match yapf codebase, but they are different to eg pep8 the linter.

### DIFF
--- a/yapf/yapflib/yapf_api.py
+++ b/yapf/yapflib/yapf_api.py
@@ -23,6 +23,7 @@ These APIs have some common arguments:
   style_config: (string) Either a style name or a path to a file that contains
     formatting style settings. If None is specified, use the default style
     as set in style.DEFAULT_STYLE_FACTORY
+  style_opts: (dict) Dictionary of style overrides.
   lines: (list of tuples of integers) A list of tuples of lines, [start, end],
     that we want to format. The lines are 1-based indexed. It can be used by
     third-party code (e.g., IDEs) when reformatting a snippet of code rather
@@ -51,6 +52,7 @@ from yapf.yapflib import subtype_assigner
 
 def FormatFile(filename,
                style_config=None,
+               style_opts=None,
                lines=None,
                print_diff=False,
                verify=True):
@@ -70,6 +72,7 @@ def FormatFile(filename,
 
   return FormatCode(original_source,
                     style_config=style_config,
+                    style_opts=style_opts,
                     filename=filename,
                     lines=lines,
                     print_diff=print_diff,
@@ -79,6 +82,7 @@ def FormatFile(filename,
 def FormatCode(unformatted_source,
                filename='<unknown>',
                style_config=None,
+               style_opts=None,
                lines=None,
                print_diff=False,
                verify=True):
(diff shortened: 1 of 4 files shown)


yapf/yapflib/yapf_api.py
```diff
@@ -51,6 +52,7 @@
 
 def FormatFile(filename,
                style_config=None,
+               style_opts=None,
```
alexjurkiewicz: oops, my editor is too helpful, the commit strips intentional trailing whitespace from a bunch of tests... I'll push a fix for this after your feedback

yapftests/yapf_test.py
```diff
@@ -491,9 +511,9 @@ def g():
 
 
 
-        #comment
```
alexjurkiewicz: oops, stripped trailing whitespace. Will correct following your feedback

eliben:
Can you explain what the benefit of this is?

alexjurkiewicz:
I work on a codebase that uses the pep8 style guide with a different line
width. yapf is really nice, and I'd like to more easily support running it
on this codebase.

More broadly, I think line width is the most common deviation from the pep8
standard, and it's nice to support people changing it in an simple fashion.
Using the `--style` dictionary approach requires reading the source code,
and even if documentation improves it requires writing a dict as a
commandline parameter, which is not super friendly.

eliben:
On Mon, Apr 13, 2015 at 6:01 AM, Alex Jurkiewicz <notifications@github.com>
wrote:

> I work on a codebase that uses the pep8 style guide with a different line
> width. yapf is really nice, and I'd like to more easily support running it
> on this codebase.
>
> More broadly, I think line width is the most common deviation from the pep8
> standard,


Different people may have different opinions here.


> and it's nice to support people changing it in an simple fashion.
> Using the `--style` dictionary approach requires reading the source code,
>

We'll gladly improve the documentation to expose this more explicitly.


> and even if documentation improves it requires writing a dict as a
> commandline parameter, which is not super friendly.
>

Why is is not friendly to point YAPF to a configuration file for your
particular style?

It's just:

  $ yapf --style <filename>

I don't see why this is too long, or too unfriendly. You write that style
file once, and you're done.

If you don't want to run long command-lines, just wrap yapf with a shell
script or an alias or whataver so you can run:

  $ myyapf

And get whatever you need.

hayd:
IMO you should either add all pep8 cli options or none. personally I prefer the current style (but I agree the available options should seen in `--help`).

One thing that pep8 does (autopep8 and pep8radius do too) is look for a config file, so you don't need to pass a style/options. This is pretty useful, and tricky/not possible to add with a simple alias.

*Though it's questionable (unfortunately) whether there's any standard for this, and unclear whether these scripts ought to have read access to random files on your system (!) in the hope they find a config file... I it's a nice enhancement/option to have.*

eliben:
I think #102 is a pretty good solution for making things transparent

PR #100, opened 2015-04-14
Local config search

hayd:
If you pass --style=local, then look up the project's config file.

This replicates the (local) behaviour of pep8/autopep8, global could work similar but I really dislike that... Note: in those libraries it is actually the default behaviour (to apply global and local configs before pep8ing).

Happy to add a bit more documentation if this was considered, or update with a different API.

cc #93

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[style]
+based_on_style = yapf
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -91,6 +91,9 @@ def main(argv):
     print('yapf {}'.format(__version__))
     return 0
 
+  if args.style[:5].lower() == 'local':
+    args.style = yapf_api.style.SearchForLocalConfig(args.files)
+
   if args.lines and len(args.files) > 1:
     parser.error('cannot use -l/--lines with more than one file')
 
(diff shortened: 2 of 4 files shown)


hayd:
squashed.

eliben:
This seems like an unusual approach. Can you explain why simply specifying a file with `--style` isn't good enough?

An alternative idea that was floating around is to have some sort of `.yapfstyle` file yapf could look for - this seems easier / more transparent than the approach suggested here.

hayd:
It does seem unusual (!)... this is the way that pep8 etc. do it. I actually don't have a compelling argument, aside from that.

However, I do think it's convenient for the user to not have to work out the location of the config, e.g. if you in a directory that's not in the project home....

`.yapfstyle` would work for me, and be much more transparent.

*Note: There's also a bad-taste here about passing dicts to CreateStyleFromConfig which I ought to clean up, but I wanted to throw this out to discuss/bikeshed the API.*

sbc100:
Ah, yes I was going do make a very similar change.

I think it should be the default though.  i.e. style default to pep8 unless a local .style.yapf file is found in a parent folder.   This is how clang-format works I think.   Also, I think the file name should start with a '.', since it fits into the category of project metadata and there is precedence for using hidden file names for this kind of thing.

sbc100:
My version of this change is at #108.  Mine one does a per-file default lookup, which I think matched the behavior of clang-format.

hayd:
+1 to that implementation.

PR #108, opened 2015-04-17
Add support for project-local style (in .style.yapf)

sbc100:
YAPF will now look for this file at startup and use it as
the default if found.  If not found pep8 is still the default.

Now that we have project-local style there is no need for the
hardcoded yapf style (it will be the default for all files
in the yapf tree now).

### DIFF
--- a/yapf/yapflib/file_resources.py
+++ b/yapf/yapflib/file_resources.py
@@ -23,6 +23,24 @@ import re
 from lib2to3.pgen2 import tokenize
 
 from yapf.yapflib import py3compat
+from yapf.yapflib import style
+
+
+def GetDefaultStyleForDir(dirname):
+  """Return default style name for a given directory.
+
+  Looks for .style.yapf in the parent directories and return the filename if
+  found, otherwise return the glboal default (pep8)."""
+  dirname = os.path.abspath(dirname)
+  while True:
+    style_file = os.path.join(dirname, style.LOCAL_STYLE)
+    if os.path.exists(style_file):
+      return style_file
+    dirname = os.path.dirname(dirname)
+    if not dirname or dirname == os.path.sep:
+      break
+
+  return style.DEFAULT_STYLE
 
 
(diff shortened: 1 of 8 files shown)


gwelymernans:
What's the status of this pull request? @eliben, do you have any issues here? I have some preference, but this area is not my specialty. So I am comfortable to defer to someone else on this. :-)

eliben:
I think both you and I left review comments for @sbc100 and are waiting for them to be addressed?

sbc100:
Yes, I have some work to do still here.  Will try to get it updated today.

sbc100:
PTAL

yapf/yapflib/file_resources.py
```diff
@@ -23,6 +23,27 @@
 from lib2to3.pgen2 import tokenize
 
 from yapf.yapflib import py3compat
+from yapf.yapflib import style
+
+
+LOCAL_STYLE = '.style.yapf'
```
hayd: Github seems to have expunged the previous comments (they're not even on the commit), but IIRC one suggestion was to move this var to style.py.

...At the moment the tests fail :(

sbc100: Sorry, I thought you were referring to DEFAULT_STYLE, which I did move.

Regarding comments, that is one of things I don't like (or don't understand) about github review system.  I pushed my updated change with push -f .. so the SHA changed, I could have made tweaks in the new change but I don't donn't to pullute history with lots tweeks even before a change lands.  Perhaps you know some other better why to do this than push -f?

hayd: If the comment is on the PR/files page they remain as "outdated diff" (even if you force push). Not sure whether the comments on google/yapf/commit  remain after a force push, but it looks like if they're on commits on your fork they are definitely not shown.

This comment was @eliben, so not sure what the intention was (I was misremembering the line).

Issue #89, opened 2015-04-12
A line can be split between unary operator `not` and its operand


gpshead:
Trying yapf out on some random pieces of code I came across this odd looking diff:

```diff
-        remote_checksum = self.get_checksum(conn, tmp, dest, inject, not direct
ory_prepended, source=source)
+        remote_checksum = self.get_checksum(conn, tmp, dest, inject, not
+                                            directory_prepended,
+                                            source=source)
```

It left the `not` on the original line but moved the operand `directory_prepended` to the next.  unary operators need to be stuck to their operands. They should always be on the same line.

(there's a separate issue regarding binary operators which can do the opposite with the operator ending up starting a line; probably a related code path but I haven't dug in to find out - https://github.com/google/yapf/issues/38)

felipevolpone:
I'm trying to fix it, but I'm having a little bit of difficulty to find where this change must be done. I'm looking in the subtype_assigner.py and in the continuation_splicer.py module, but I'm not sure of how the splice of each Node is done. Furthermore, I guess that it would be necessary check if the current node is a NOT Node and if the next node is a valuable Node. 

If there is some one interested and with time to help me, I would like to do this change and help the project. Thanks

gwelymernans:
@felipevolpone Sorry for the delay in getting back to you. The problem is that the penalty for splitting after the `not` is not high enough to prohibit splitting before it. You might want to look at `Visit_not_test` in `split_penalty.py` to see if adjusting the `NOT_TEST` value helps out here. The `split_penalty.py` file exists to add penalties for things we can derive from the `pytree` itself.

There is another option, which is to adjust the penalty in `_SplitPenalty()` in `unwrapped_line.py`. The reason for this is to apply penalties outside of the `pytree` structure, and takes only the current and previous nodes into account. So it's a bit more general than the first, but offers an easy way to adjust the penalties for certain constructs. If you do it there, then consider making it `split_penalty.STRONGLY_CONNECTED` or `split_penalty.UNBREAKABLE`.

felipevolpone:
Hey @gwelymernans , really thanks for your help and your time. I send a pull-request trying to solve this. You can check https://github.com/google/yapf/pull/114 . Maybe I not have done in the best way, but you can tell me this too.

gwelymernans:
Thanks for the patch, @felipevolpone! I added a few comments to it.

PR #115, opened 2015-04-20
Resolving #89: A line can be split between unary operator `not` and its operand 

felipevolpone:
I changed the unwrapped_line._SplitPenalty method to check if the not operator is followed by True or False. Also, I created a test called split_operator_not.py to show the fix. Really thanks to @gwelymernans who helped me to do this. Maybe I not have done in the best way, but you can tell me this.

Tks


felipevolpone:
Hey @gwelymernans, I'll do the changes that you suggested and create another PR. Ok?

felipevolpone:
I'll create another pull request to do these changes.

PR #120, opened 2015-04-21
Add check cli option

petedmarsh:
The --check option determines if the given files adhere to the specified style or not. If none of the files require reformatting it will exit normally and log nothing to STDOUT. If reformatting is required to meet the standard then it will print the diff and exit with status 3.

A couple of things I should note:

1. This prints the diffs of files that need reformatting (if there are any), as such there's a certain amount of overlap with the --diff option. Maybe it should output something different, like the out-of-style file paths?
2. I chose exit status 3 relatively arbitrarily, it's the first non-reserved exit status
3. I have not formatted the code, I ran `yapf --style chromium --diff -r .` before I started working on this and noticed some files that need formatting, I thought I'd keep the diff small :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -86,6 +86,8 @@ Options::
       --style STYLE         specify formatting style: either a style name (for
                             example "pep8" or "google"), or the name of a file
                             with style settings. pep8 is the default.
+      -c, --check           heck the files adhere to the style, exits with
+                            status 3 if not
       -d, --diff            print the diff for the fixed source
       -i, --in-place        make changes to files in place
       -l START-END, --lines START-END
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -73,13 +73,19 @@ def main(argv):
   parser.add_argument('--verify',
                       action='store_true',
                       help='try to verify refomatted code for syntax errors')
-  diff_inplace_group = parser.add_mutually_exclusive_group()
-  diff_inplace_group.add_argument('-d', '--diff',
-                                  action='store_true',
-                                  help='print the diff for the fixed source')
-  diff_inplace_group.add_argument('-i', '--in-place',
-                                  action='store_true',
-                                  help='make changes to files in place')
+  diff_inplace_check_group = parser.add_mutually_exclusive_group()
+  diff_inplace_check_group.add_argument('-c', '--check',
+                                        action='store_true',
+                                        help=('check the files adhere to the '
+                                              'style, exits with status 3 if '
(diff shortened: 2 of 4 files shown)


eliben:
Out of curiosity, can you clarify why this is needed?

For the odd project that needs such functionality, it sounds like a very trivial wrapper around yapf can do it.

hayd:
You can use `test -n` (to test whether the diff is non-empty).

    yapf yapf --style=yapf --diff | test -n -

You could do a similar check in python, and you could make it fail early (e.g. with any).

IMO most of the time you won't want the diff when doing a check like this...

petedmarsh:
It's really for making something like a CI build fail if the source code is in the wrong style - I think you're right @eliben that it's easy to get this behavior if you want it (as @hayd has shown). 

Happy to close this PR - however I think 1e8ecc878bafa30a23bd068409d408a16cc7c57e fixes a genuine bug, is that the case? If so I'll raise a separate PR with just that commit.

eliben:
Yes, closing this one and opening a separate PR for the bug fix sounds good

hayd:
I'm not sure that that's a bug, the empty diff is `''` (not None).

That said. It might be nice to add the "testing for satisfying yapf" to the readme, as I agree people will want to do that for sure. No idea if there's a windows equivalent to test... pipe it to python?

petedmarsh:
I'm not sure 100% about the empty diff being `''` vs. None too, I noticed that `FormatFiles` (https://github.com/google/yapf/blob/master/yapf/__init__.py#L172) checks if the result of `apf_api.FormatFile` is `None`, *but* from what I could see it never can be None, unless the empty diff was supposed to be `None` - either way all the tests pass do it's hardly a huge deal! :)

hayd:
Hmmm, I think maybe that can be None [due to this section](https://github.com/google/yapf/blob/1e8ecc878bafa30a23bd068409d408a16cc7c57e/yapf/yapflib/yapf_api.py#L69-L71). *(The alternative may be to change the [`reformatted_code  is not None`](https://github.com/google/yapf/blob/1e8ecc878bafa30a23bd068409d408a16cc7c57e/yapf/__init__.py#L178) to `if reformatted_code`, but I don't think that changes behaviour...)*

gwelymernans:
There is the degenerate case of the file being nothing be whitespace. We would want to replace that file with `''`, I think. However, because of the `if not reformatted_code.rstrip(): return` in the function that writes to file, the file won't change. Which is probably a bug...

Issue #105, opened 2015-04-15
Splitting function arguments into new lines when column limit not reached


sawan:
Turned this:
```python
 def main():
     p = ArgumentParser(add_help=False)
-    p.add_argument('--config', default=CONFIG,
-                       help='Filename of config file (default %(default)s).')
-    p.add_argument('--interval', type=int, default=INTERVAL,
-                                  help='Cycle time for script (default %(default)d).')
-    p.add_argument('--offset', type=int, default=OFFSET,
-                       help='Delay time for script cycle (default %(default)d).')
-    p.add_argument('--connect-string', default=CONNECT_STRING,
-                       help='Database connect string (%(default)s).')
-    p.add_argument('--proc', default=PROC,
-                       help='SQL Procedure  (%(default)s)')
```
 into:
```python
+    p.add_argument('--config',
+                   default=CONFIG,
+                   help='Filename of config file (default %(default)s).')
+    p.add_argument('--interval',
+                   type=int,
+                   default=INTERVAL,
+                   help='Cycle time for script (default %(default)d).')
+    p.add_argument(
+        '--offset',
+        type=int,
+        default=OFFSET,
+        help='Delay time for script cycle (default %(default)d).')
+    p.add_argument('--connect-string',
+                   default=STRING,
+                   help='Database connect string (%(default)s).')
+    p.add_argument('-proc',
+                   default=PROC,
+                   help='SQL Procedure  (%(default)s)')
```

My query/problem here is: why have some of the arguments been split off into another line when they would be well within the column limit if left on the same line? Is there a config option I can use?

hayd:
I don't see this behaviour in master (all arguments line up in the same way).

Are you using a style with COLUMN_LIMIT less than 79?

gwelymernans:
These were split because we tend to push towards one item per line when the function call has "keyword=value" pairs in it. I should probably add an option for this...

sawan:
@hayd I have not changed any defaults.

@gwelymernans Yes, I think that is needed.

gwelymernans:
I forgot that we already have an option for this. It's the `SPLIT_BEFORE_NAMED_ASSIGNS` option. If you disable it, then it should give you something more reasonable. :-)

sawan:
Tried with:

```
--style='{SPLIT_BEFORE_NAMED_ASSIGNS=False} AND
--style='{SPLIT_BEFORE_NAMED_ASSIGNS=True}'
```

made no difference.

hayd:
[...]

PR #122, opened 2015-04-22
Be less strict with dict-style input.

hayd:
Accept {INDENT_WIDTH=4} or {INDENT_WIDTH : 2}
Raise StyleConfigError if given bad input.

cc #105 @sawan.


hayd:
Of course, perhaps the better options is to be more strict...

gwelymernans:
It's really meant to be a "nicety" for the user. Given that, making it less strict isn't a bad idea I suppose.

PR #125, opened 2015-04-22
Fix edge case of python file which is blank lines.

hayd:
Before this was just passed through (no diff was shown, nothing happened inplace).

See https://github.com/google/yapf/pull/120#issuecomment-95316601.

### DIFF
--- a/yapf/yapflib/file_resources.py
+++ b/yapf/yapflib/file_resources.py
@@ -39,8 +39,6 @@ def WriteReformattedCode(filename, reformatted_code, in_place, encoding):
     in_place: (bool) If True, then write the reformatted code to the file.
     encoding: (unicode) The encoding of the file.
   """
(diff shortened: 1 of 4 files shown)


hayd:
Not clear to me why this fails on python 2. :(

gwelymernans:
I think the reason why is because it looks like an empty string is considered a `str` in Python 2.7. If I wrap it with `unicode`, then it works. It's probably because of the encoding that the file is opened with.

hayd:
@gwelymernans But the error is:

    AssertionError: u'' != u'\n'

?

gwelymernans:
Ah! I see. I'm not sure what to say. Try printing out values right before it writes to the file?

hayd:
Ha! python 2 was sensitive about unicode. One line fix...

*Strangely the raised exception was not the reported fail.*

PR #127, opened 2015-04-24
#issue89: line can be split between unary operator not

felipevolpone:
This pull request refers to the https://github.com/google/yapf/pull/115 pull request. I just changed a little based on the tips of @gwelymernans .


felipevolpone:
@gwelymernans  Can you please check again?

gwelymernans:
@felipevolpone I'm sorry for the delay. I still don't think this is the correct approach. You're checking for `not` as the previous token and `True` or `False` for the current token. But that won't work for all situations. I don't think you need to check that the current token is anything, only that the previous token is `not`.

yapf/yapflib/unwrapped_line.py
```diff
@@ -383,6 +383,11 @@ def _IsSurroundedByBrackets(tok):
 
 def _SplitPenalty(prev_token, cur_token):
   """Return the penalty for breaking the line before the current token."""
+
+  if prev_token.value == 'not' and (cur_token.value == 'True'
```
hayd: I can confirm that this passes your tests as simply:

    if prev_token.value == 'not':

change this, squash and rebase. I think this is a nice enhancement!

yapftests/reformatter_test.py
```diff
@@ -1914,6 +1914,30 @@ def testExecAsNonKeyword(self):
     self.assertEqual(expected_formatted_code, reformatter.Reformat(uwlines))
 
 
+class TestingNotInParameters(unittest.TestCase):
+
+    def test_notInParams(self):
+        unformatted_code = textwrap.dedent("""\
+            def sum(a, sprint):
```
hayd: (Though I don't like your test overriding the builtin sum function :p these tests would work just as well with just the last line.... but I don't think this matters.)

felipevolpone:
@gwelymernans  and @hayd: thanks for the help. 
Based on what @hayd said I created another pr #149. If you could see and give some feedback, it would be great.

hayd:
FWIW you can "force push" changes to a branch and it'll update the pull-request (no need to do it this time but perhaps good to know).

gwelymernans:
This seems to be done now with #149. Thanks!

PR #136, opened 2015-05-06
Add argument `--stdout-encoding`

wklken:
When use yapf to format code from stdin

Cannot choose the stdout encoding(my stdout encoding is `utf-8`), and then
```
#!/usr/bin/env python
# encoding: utf-8

# 中文
```

the plugins which use `yapf` 
will case error

```
  File "/usr/local/bin/yapf", line 11, in <module>
    sys.exit(run_main())
  File "/usr/local/lib/python2.7/site-packages/yapf/__init__.py", line 222, in run_main
    sys.exit(main(sys.argv))
  File "/usr/local/lib/python2.7/site-packages/yapf/__init__.py", line 142, in main
    verify=args.verify))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 43-44: ordinal not in range(128)
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -94,6 +94,8 @@ Options::
                             directories of the source file (or current directory
                             for stdin)
       --verify              try to verify refomatted code for syntax errors
+      --stdout-encoding STDOUT_ENCODING
+                            define stdout encoding
       -d, --diff            print the diff for the fixed source
       -i, --in-place        make changes to files in place
       -l START-END, --lines START-END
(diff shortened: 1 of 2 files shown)


wklken:
I signed it!

gwelymernans:
Please add a testcase to `yapf_test.py` and squash all of the commits into one.

Actually, what I would like to see more is detecting the encoding via the `tokenizer.detect_encoding` method. Similar to `ReadFile` in `yapf_api.py`. Could you look into doing that?

wklken:
ok. I will find out how to detect encoding in stdin/stdout, change the code and add a testcase, 
recommit later:)

Issue #139, opened 2015-05-07
Block comments should have a space after # in pep8 style mode


MarkusTeufelberger:
https://www.python.org/dev/peps/pep-0008/#block-comments - the pep8 checker throws E265 on a test file that only contains `#This is a comment`

Expected behaviour would be to insert a space character after the # character, producing `# This is a comment\n`.

gwelymernans:
Formatting of comments is a completely separate issue, and more complex than just adding a space after the `#`. There are many things to take into consideration, including comments that are meant to be this way (e.g., internationalization comments), comments that will go over the column limit when the space is added, the `#!` comments, etc.

Given that, I think that it should be optional. Possibly done via a `lib2to3` "fixer".

MarkusTeufelberger:
I would expect that the following workflow works once yapf is released as 1.0:

1. Write (badly formatted) code
2. run pylint/flake8/... to check for pep8 conformity --> errors are found
3. run yapf in pep8 mode on the code
4. run pylint/flake8/... again --> only errors that can not be fixed by a simple formatting change (e.g. renaming variables beyond just changing between UPPER and lower case or inserting_underscores, McCabe complexity warnings...) are left
5. ???
6. profit!

I don't think there are any edge cases where spaces are not allowed beyond shebang lines (#!/usr/bin/env python), and these might even be useful if they could be automatically added as first line in a file, especially for python3.

Comments overflowing the 80 (actually 79 - https://www.python.org/dev/peps/pep-0008/#id14) column limit can be just wrapped around to the next line. For longer comments or documentation I'd prefer a setting that regulates how many comment lines are considered a "long" comment, e.g. more than 3 comment lines after each other which has to be wrapped at 72 characters.

Where is the documentation for these "internationalization comments"? I coudn't find anything about special comments in https://docs.python.org/3/library/i18n.html or https://docs.python.org/2/library/gettext.html

slel:
Also requested at

- [Stack Overflow question 53527882: Spacing on inline comments with yapf](https://stackoverflow.com/q/53527882)

kamahen:
Vim has a reflow paragraph command.

And emacs, of course, also has a reflow command (also called "paragraph fill").

PR #150, opened 2015-05-27
Fixing #issue139

felipevolpone:
I'm doing the #139 improvement. I think that I made in the right place, but please tell me if I did something wrong.

### DIFF
--- a/yapftests/yapf_test.py
+++ b/yapftests/yapf_test.py
@@ -573,9 +573,9 @@ class CommandLineTest(unittest.TestCase):
 
 
 
-        #comment
+        # comment
 
-        #   trailing whitespace
+        # trailing whitespace
         """)
     self.assertYapfReformats(unformatted_code, expected_formatted_code,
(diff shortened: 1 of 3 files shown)


felipevolpone:
@gwelymernans I think that is right now.

yapftests/yapf_test.py
```diff
@@ -573,9 +573,9 @@ def g():
 
 
 
-        #comment
```
hayd: I thought the point of these not changing is that they are outside of the range 3 3.

Presumably this means that your change is not checking for being within the range(s).

bhack:
Any news on this?

govjess:
The hardest part is channeling the money where it doesn't get hacked in the process by anyone who can socialize with me online. Making a paper billfold is mandatory and not to give keys online at anytime to the private wallet.

gwelymernans:
I think the fix this PR is meant to handle already exists now.

PR #156, opened 2015-06-15
fix for infinite loop on style file search on windows

Laharah:
`os.path.dirname('C:\\')` returns `'C:\\'` on windows system, not the widows path separator `'\\'`. This prevents the break from being called and creates an infinite loop on windows file systems if there is no style file in any parent directory, making yapf hang indefinitely.

Calling `os.path.abspath(os.path.sep)` returns `'/'` on posix systems and `'C:\\'` or whatever drive letter you're on for windows systems. 

Quick `abspath` in the break-case check fixes the infinite loop.


Laharah:
Signed

PR #157, opened 2015-06-16
Add how to use as module to the README

### LINKED ISSUE #148: Examples for using as a library
Love this project. Can we get API docs for using yapf as a library please?

hayd:
Fix #148.

Note: IMO the FormatFile api is a little strange atm, in returning (source, encoding) or (diff, encoding). I think an in_place argument would be good... but I'm not sure what it ought to return in this case ([None](https://github.com/google/yapf/blob/25de78025392360224e7ec40a90c1502377e34d5/yapf/yapflib/yapf_api.py#L71), encoding) ?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -175,6 +175,65 @@ and reformat it into:
     def f(a):
         return 37 + -+a[42 - x:y ** 3]
 
+Example as a module
+===================
+
+The two main APIs for calling yapf are ``FormatCode`` and ``FormatFile``:
+
+.. code-block:: python
+
+    from yapf.yapf_api import FormatCode  # reformat a string of code
+    from yapf.yapf_api import FormatFile  # reformat a file
+    
+    >>> FormatCode("f ( a = 1, b = 2 )")
+    'f(a=1, b=2)\n'
+
+You can also pass a ``style_config``: Either a style name or a path to a file that contains
+formatting style settings. If None is specified, use the default style
+as set in ``style.DEFAULT_STYLE_FACTORY``.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -175,6 +175,65 @@ and reformat it into:
...
+You can also pass a ``style_config``: Either a style name or a path to a file that contains
+formatting style settings. If None is specified, use the default style
+as set in ``style.DEFAULT_STYLE_FACTORY``.
+
+.. code-block:: python
+
+    >>> FormatCode("def g():\n  return True", style_config='pep8')
+    'def g():\n    return True\n'
+
+    >>> FormatFile("def g():\n  return True", style_config='pep8')
```
hayd: eurgh, this is a typo. Doh!

gwelymernans:
Oops! I didn't catch this. I mentioned in the other pull request that it should probably return `None, encoding`. That seems the most logical thing to do, IMHO. But I could be wrong. :-)

README.rst
gwelymernans: Go ahead and fix it and then squash the result. :-)

hayd:
I may cheekily wait for #158 and then update to include mention of in_place! :)

Like I said there, having looked a little, I'm not *sure* that (None, encoding) makes sense, because then you have the same result whether:

- the file exists, is reformatted, writes out sucessfully
- the file does not exist, and exits\*

\* for some reason currently this returns (None, encoding) rather than raising an exception of some sort... :s

Issue #159, opened 2015-06-17
yapf freezes on windows


shelper:
I just installed YAPF on windows Python 3.4.2, via pip and try to run it on a file `foo.py` by simply put:
`yapf foo.py`, the file is as simple as one line code ` x = y+z ` and i found it freezes my computer, and never finish, 
but when i import yapf and uses the function FileFormat() it worked. I believe there is some bug, maybe windows specific, in the code？

shelper:
i found the fix here 
https://github.com/Laharah/yapf/commit/3e7937268a7142f669cbd4d8e9789d2621bb4452

dont know why the author didnt file a pull request but hopefully this can be fixed in the next release

gwelymernans:
Sorry for the delay in getting to this. I applied a patch similar to @Laharah's patch. Please try it out!

PR #160, opened 2015-06-17
- fixed infinite regression for .style search on windows filesystems

shelper:
As explained in the title, i encountered an issue 
https://github.com/google/yapf/issues/159
and the branch from Laharah fixes the issue, so i filed a pull request


hayd:
fixed by #156.

PR #161, opened 2015-06-17
Mention FormatFile's in_place argument in readme.

hayd:
Also fix a typo from updating the README in #157.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -178,26 +178,23 @@ and reformat it into:
 Example as a module
 ===================
 
-The two main APIs for calling yapf are ``FormatCode`` and ``FormatFile``:
+The two main APIs for calling yapf are ``FormatCode`` and ``FormatFile``, these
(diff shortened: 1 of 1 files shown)


hayd:
Aside: Something that may be worth doing, is having FormatFiles in `yapf_api` too... IMO having FormatFile and FormatCode available at the toplevel (i.e. as `yapf.FormatCode`) would be nice, but this is a style/subjective thing.

Ooh, that merge was quick. Thanks!

gwelymernans:
I intentionally split them into `yapf_lib`, because I felt like they were the lowest level API functions for the project. I'm iffy on `FormatFiles being in there. For instance, a user could have their own version of `FormatFiles` that returns data and then does a diff. Or calls `FormatFile` with different styles. Etc. I know these may not be actual use cases, but I wanted to be flexible.

Does that make sense? :-)

PR #179, opened 2015-08-24
Detect gyp files as python files

till-varoquaux:
Gyp file syntax is purely a subset of python's so it makes sense to use YAPF to reformat them too.

### DIFF
--- a/yapf/yapflib/file_resources.py
+++ b/yapf/yapflib/file_resources.py
@@ -98,7 +98,7 @@ def _FindPythonFiles(filenames, recursive):
 
 def IsPythonFile(filename):
   """Return True if filename is a Python file."""
-  if os.path.splitext(filename)[1] == '.py':
+  if os.path.splitext(filename)[1] in ['.py', '.gyp', '.gypi']:
     return True
 
   try:
--- a/yapftests/file_resources_test.py
+++ b/yapftests/file_resources_test.py
@@ -132,6 +132,12 @@ class IsPythonFileTest(unittest.TestCase):
     file1 = os.path.join(self.test_tmpdir, 'testfile1.py')
     self.assertTrue(file_resources.IsPythonFile(file1))
 
+  def test_with_gyp_extenstion(self):
+    file1 = os.path.join(self.test_tmpdir, 'testfile1.gyp')
+    self.assertTrue(file_resources.IsPythonFile(file1))
+    file1 = os.path.join(self.test_tmpdir, 'testfile1.gypi')
+    self.assertTrue(file_resources.IsPythonFile(file1))
+
   def test_empty_without_py_extension(self):
     file1 = os.path.join(self.test_tmpdir, 'testfile1')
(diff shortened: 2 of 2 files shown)


till-varoquaux:
I signed it

gwelymernans:
This fell off of my plate. I'm sorry about that!

Before I incorporate this, we recently modified how files are chosen for reformatting. The `IsPythonFile()` function is used only when the `--recursive` flag is set. Does that change cover this use case?

cclauss:
Bump on this one

gwelymernans:
I'm not a big fan of this approach. Many file types have Pythonesque syntaxes, and we should have a better way to specify that they should be treated as Python than hard-coding them in a list. Would a command line flag be better? Or perhaps adding it to the `.style.yapf` file?

yapf/yapflib/file_resources.py
```diff
@@ -98,7 +98,7 @@ def _FindPythonFiles(filenames, recursive):
 
 def IsPythonFile(filename):
   """Return True if filename is a Python file."""
-  if os.path.splitext(filename)[1] == '.py':
+  if os.path.splitext(filename)[1] in ['.py', '.gyp', '.gypi']:
```
siddharths2710, line 101: @gwelymernans may I suggest a different approach altogether? Instead of checking the file type, we can use the `ast` syntax tree module to check if the content is Python compatible. The [parser](https://docs.python.org/3/library/ast.html#ast.parse) raises a `SyntaxError`  if the code isn't Pythonic.

kamahen: You probably want to add `'.pyi'` to the list. (It'll also work with the `ast` check)

One problem with the `ast` check is that a Python file with a syntax error will just be skipped. Currently, it produces a rather unattractive crash. Whether this is a bug or a feature depends on the use case -- if an editor calls yapf (e.g., using emacs' flymake), it could be useful to mark the syntax error rather than silently do nothing.

cclauss: Use ['.py', '.gyp', '.gypi'] instead of ['.py', '.gyp', '.gypi'] because:
1. tuples are smaller and more efficient than lists
2. a tuple signals to the reader that the contents are fixed and can not be modified at runtime.

I am not a fan of the ast approach for performance reasons and because it would not alert users of malformed Python files.

gwelymernans: I agree with @cclauss. Parsing the file more than once will make things slower than need be. We already have performance problems, because the base algorithm is O(N^2). It could also lead to yapf trying to modify a file that isn't meant to be formatted (i.e., it's not a Python file, but the `ast` parser parses it without error).

Another issue is that if we hardcode the file extensions, then we'll have to change the list for any future additions, which is kinda annoying and makes me think that there should be a Better Way(tm). I suggested adding it as an option to .yapf.style, or command line parameter, etc., for that reason. I'm not saying that we need to do it that way, but I would like a more flexible option.

PR #185, opened 2015-09-13
Fix formatting of multiple files

joaodelgado:
FormatFiles was returning after formatting the first file, even if multiple files were provided.


gwelymernans:
Doh! Thank you for the fix. Sorry for the oversight.

PR #187, opened 2015-09-22
Add DEDENT_CLOSING_BRACKETS

ambv:
This allows for PEP8-compatible JSON-style scope organization that has the
following properties:

* produces smaller diffs
* uniformly treats all closing brackets, including function definitions,
  calls and data structures
* leaves more characters per line by solving the PEP8 violation known as
  "E125: continuation line with same indent as next logical line" without
  additional indentation necessary

## Some remarks
* This mode fixes a number of "yapf: disable" necessary in yapf's own codebase
* Performance might not be awesome since we're searching opening brackets on-the-fly. I'll improve that if needed.
* Tox on Python 2.7 and Python 3.4 passes but I didn't cover this with new tests yet. I'll create them when you "A-OK" the idea.
* I bumped the version according to semver
* I added a "facebook" style that closely resembles what we endorse internally
* Oh, and most importantly &mdash; thanks for creating this tool! It's way better than autopep8 and others we've unsuccessfully tested in the past

### DIFF
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -38,7 +38,7 @@ from yapf.yapflib import py3compat
 from yapf.yapflib import style
 from yapf.yapflib import yapf_api
 
-__version__ = '0.3.0'
+__version__ = '0.3.1'
 
 
 def main(argv):
--- a/yapf/yapflib/reformatter.py
+++ b/yapf/yapflib/reformatter.py
@@ -112,14 +112,13 @@ def _RetainVerticalSpacing(cur_uwline, prev_uwline):
   prev_tok = None
   if prev_uwline is not None:
     prev_tok = prev_uwline.last
-  cur_lineno = 0
-  prev_lineno = 0
   for cur_tok in cur_uwline.tokens:
     _RetainVerticalSpacingBetweenTokens(cur_tok, prev_tok)
     prev_tok = cur_tok
 
 
 def _RetainVerticalSpacingBetweenTokens(cur_tok, prev_tok):
+  """Retain vertical spacing between two tokens."""
   if prev_tok is None:
     return
 


ambv:
Oh, as for the CLA, I think Facebook has some corporate CLA with you, don't we?

ambv:
Alright, I submitted a CLA for my GitHub account.

ambv:
Added some tests but more to come (see remarks for under-reported coverage). There's one missing piece I'd like to address before landing this (see TODOs I added in `yapf_test.py` for details).

Some general observations:

* the current `--style=pep8` format is not PEP8 compliant (see TODOs I added in `yapf_test.py` for details); might be worth tackling in a separate change
* the unit tests reveal that the `style` design is flawed with it being global shared state; Eli is right here, it should rather be passed as an argument across the API
* there’s a lack of integration tests at the moment; yapf_test.py is a start but it’s not comprehensive yet and using subprocess hinders perf a lot and underreports coverage. The way the different components interact is subtle and for now we’re not tracking regressions. Integration tests could for instance help tackling user-perceived inconsistency in the generated style (I will file a separate issue for this)

Nits:

* yapf/yapf/yapflib is a strange and mildly annoying path to have to write; especially with yapf/yapftests/ tab-conflicting; yapf/src/yapflib/ and yapf/tests/ would be nicer. This is also seen in imports: “from yapf.yapflib” is needless repeating yourself.

gwelymernans:
* I think you're right that tackling the PEP8 issue you saw in a separate pull request would be the best way to go. It will most likely be a more invasive change.

* I'm with you on the `style` issue. It would be nice to pass it around where needed. The way it is now is due to Historical Reasons(tm).

* The naming is...well weird, I know. I was actually following how `rope` arranged their directories. I know that it's a bit stutter-y, but it kind of fits our internal naming style, which suffers a bit from echolalia. That said, I'm open to suggestions ...

hayd:
RE tests: [autopep8 has a rather large suite](https://github.com/hhatto/autopep8/tree/master/test), and I think they do some acid/random testing on subsets of python core libs. IIRC They mostly check that the ASTs remain the same (and perhaps that pep8 returns non-zero on the fixed code).

gwelymernans:
Yeah, `autopep8` has a much more extensive test suite than we do. But they're also testing for different things. :-) They check that the transformation is correct by doing a "verification" run; essentially making sure the code compiles. We tried to do that, but ran up against a ton of problems with it. You can still use it with the `--verify` flag. I don't know about them running it against `pep8`, but that might not be a bad idea for us to do, at least for testing...

hayd:
I thought that autopep8 did AST-preserving transformations (if *not* using `--aggressive`), and they tested whether these ASTs were indeed unchanged by autopep8  (which is different from `--verify`ing for syntax errors) on a suite of python files. Maybe I was misunderstanding the verification step.

ambv:
Oops, this closed because I failed at git and rebasing. The changes are still there on https://github.com/ambv/yapf/tree/dedent. Will recreate the pull request (mentioning this one) when I'm done with test coverage.

gwelymernans:
LOL! I'm bad at rebasing too. Let me know when it's ready!

gwelymernans:
@hayd I could just be forgetting something...or I look at the wrong type of verification in `autopep8`. Anyway, it's something that would be great to have here, but we ran into compliation troubles when we tried it. :-(

hayd:
cc @myint who knows more about this than me. :)

myint:
The `autopep8` acid test is where we compile the result to check for syntax errors. Given a set of input files, the test runs `autopep8` over them. It then checks if the outputs [compile](https://github.com/hhatto/autopep8/blob/master/test/acid.py#L104). It then runs `pep8` over the files and issues warnings to standard error if there are still `pep8` violations.

https://github.com/hhatto/autopep8/blob/master/test/acid.py

There is also a [byte code comparison](https://github.com/hhatto/autopep8/blob/master/test/acid.py#L162) that can be optionally enabled, but I've found that to produce false positives.

PR #191, opened 2015-10-06
Add DEDENT_CLOSING_BRACKETS

ambv:
This allows for PEP8-compatible JSON-style scope organization that has the
following properties:

* produces smaller diffs
* uniformly treats all closing brackets, including function definitions,
  calls and data structures
* leaves more space on the line by solving the PEP8 violation known as
  "E125: continuation line with same indent as next logical line" without additional indentation necessary

## Summary of the changes in this diff
* this is an updated version of #187 which was closed by mistake
* adds `DEDENT_CLOSING_BRACKETS` (`False` by default)
* I added a "facebook" style that closely resembles what we endorse internally
* `_DetermineMustSplitAnnotation` is ran on imports, except clauses, function definitions and calls
* I bumped the version according to semver

### DEDENT_CLOSING_BRACKETS

When DEDENT_CLOSING_BRACKETS decides to put a break before the closing bracket, we are now forcing a break after the **opening** bracket as well. This is done both to have more horizontal space for the trailer and for consistency. By consistency I mean both the resulting code and internal consistency within YAPF which docstrings say in format_decision_state.py that we only want to split_before_closing_bracket if there was a newline after the opening one.

### pytree_unwrapper._DetermineMustSplitAnnotation
This worked only for lists, dicts and sets. I made it work for `import_as_names`, `testlist_gexp`, `arglist` and `typedargslist` as well. Again, consistency. This needed one change in your tests (first example in `testAlignClosingBracketWithVisualIndentation`) but I think the current formatting is better. If you don’t like this being the default, we can add a toggle for it.

### Tests
* added TestsForFBStyle in reformatter_test.py to ensure coverage

## Future design remarks
Those observations shouldn't block merging of this PR. They're rather opening the discussion about where do you want YAPF to go. 
### _DetermineMustSplitAnnotation feels arbitrary

Granted, the comma at the end or the presence of comments are nice heuristics. However, this produces vastly different results based only on the presence of one semantically insignificant character at the end of the container literal.

Instead, it would be clearer if we always preferred single-line lists when they fit. The high-level design would be like this:
* always prefer single-line formatting if it fits
* then prefer a single-line **element list** (argument list, etc.) if it fits
* otherwise, split one element per line

That’d be very easy to explain, especially seeing an example like this:
```python
# the following fits in one line
singleline(a, b, c)

# the following argslist fits in one line
singleline_separate_args(
  a, b, c
)

# there’s no way to fit the entire thing
# or its argslist in one line
multiline(
  a,
  b,
  c
)
```
Then the trailing comma becomes insignificant from YAPF’s perspective.

### Lack of idempotency
I noticed some of the reformattings are not idempotent, e.g. if you do multiple passes, the result will be mutating a couple of times. It will land at some equilibrium but it shows that some heuristics went a little too far (I suspect those that look at if there's a newline in the original source).
### Performance
There is a noticable performance hit because of _DetermineMustSplitAnnotation being run more often but no comprehensive benchmarks have been made.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,10 @@
 # All notable changes to this project will be documented in this file.
 # This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.4.0] UNRELEASED
+### Added
+- Support for dedenting closing brackets, "facebook" style.
+
 ## [0.3.2] 2015-10-04
 ### Fixed
 - Formatting of tokens after a multiline string didn't retain their horizontal
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -12,3 +12,4 @@
 Bill Wendling <morbo@google.com>
 Eli Bendersky <eliben@google.com>
 Sam Clegg <sbc@google.com>
+Łukasz Langa <ambv@fb.com>
--- a/README.rst
+++ b/README.rst
@@ -184,7 +184,7 @@ share several arguments which are described below:
 .. code-block:: python
 
     >>> from yapf.yapf_api import FormatCode  # reformat a string of code
-    
+
     >>> FormatCode("f ( a = 1, b = 2 )")
     'f(a=1, b=2)\n'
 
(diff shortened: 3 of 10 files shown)


ambv:
I updated the PR with both nits fixed.

gwelymernans:
Thank you!

PR #207, opened 2015-11-05
Update Readme to use correct filename for local style

JMoravec:
According to https://github.com/google/yapf/blob/master/yapf/yapflib/style.py#L368 the filename in the readme is incorrect. This simply changes it to use the correct filename.


JMoravec:
I've signed the CLA.

gwelymernans:
Good catch. Thanks!

Issue #224, opened 2016-01-14
Infer indentation level when formatting from stdin


strokirk:
While coding in vim, it's very useful to set yapf as the `equalprg`, which means that selected lines will be sent via `stdin` to yapf and replaced with it's output.

However, this means that if I want to format only a few lines in the middle of an indented block, yapf will fail with `IndentationError: unexpected indent`. For example, only sending the `bar = sorted( bar )` line to yapf in the following example.

```python
def foo():
    bar = ['c', 'd', 'a']

    bar = sorted( bar )

    return foo
```

It would be really useful if yapf could look at the first line of input to infer the common indent-level of all following lines.

bolov:
+1

gwelymernans:
Unfortunately, this isn't something we can control. The parser we use (`lib2to3`), isn't able to accept this. The best you can do here is use the `--lines` option when formatting in `vim`.

purpleP:
@gwelymernans Well, you can try other parser (Jedi for example) or dedent lines before sending it to parser.

gwelymernans:
@purpleP two things:

1. Dedenting the code may allow the line to go over the max column limit. If anyone wanted to go down this route, they would need to modify the max column limit amount during that formatting. If someone wanted to submit a pull request with this feature, I would consider it. :-) Then again, when the code is submitted via `stdin`, why not dedent it before, pass in the adjusted column limit, and then indent the result themselves?
2. Using a new parser, while it would have some benefits, would require a huge rewrite of the entire formatter, and then only to support one feature that has a workaround (that I mentioned above). The only way to achieve the same functionality as the Jedi parser would probably be to modify `lib2to3` to have it.

purpleP:
@strokirk @bolov  Hi, I've added pull request to fix this behaviour. Until it's merged you can use my fork if you want.

purpleP:
@gwelymernans Oh, I haven't thought about column limit. I will fix this in my pull request.

purpleP:
@gwelymernans Okay, I've added the test and fixed it. By the way, you should consider using pytest instead of unittest, because it's really much better. After using it for about half year I wouldn't want to come back to unittest even if someone would pay me for that.

PR #243, opened 2016-04-22
Support for formatting partial blocks of codes.

purpleP:
This as far as I've tested closes [https://github.com/google/yapf/issues/224](#224)
I've ran your tests (changing two of them which were checking formatting of bad input) they all green.
Also I've tested if manually on some inputs and it works fine.
One thing though. If the indentation was incorrect (not a multiple of four etc) yapf wouldn't fix it for partial inputs. Which is actually easy to fix. But me or someone else could add this later.

### DIFF
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -212,13 +212,13 @@ def FormatFiles(filenames,
           print_diff=print_diff,
           verify=verify,
           logger=logging.warning)
+      if reformatted_code is not None:
+        file_resources.WriteReformattedCode(filename, reformatted_code, in_place,
+                                            encoding)
       changed |= has_change
     except SyntaxError as e:
       e.filename = filename
       raise
-    if reformatted_code is not None:
-      file_resources.WriteReformattedCode(filename, reformatted_code, in_place,
-                                          encoding)
   return changed
 
 
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -30,7 +30,12 @@ from __future__ import print_function
 import argparse
 import logging
 import os
-import sys
+from re import match
+
+try:
+    from textwrap import dedent, indent
+except ImportError:
+    from py3compat import indent
 
 from yapf.yapflib import errors
 from yapf.yapflib import file_resources
@@ -38,8 +43,10 @@ from yapf.yapflib import py3compat
 from yapf.yapflib import style
 from yapf.yapflib import yapf_api
 
-__version__ = '0.7.1'
+__version__ = '0.6.3'
 
+
+import sys
 
 def main(argv):
   """Main program.
@@ -143,16 +150,26 @@ def main(argv):
         original_source.append(py3compat.raw_input())
       except EOFError:
         break
+
+    original_source.append('')
+    text_to_format = '\n'.join(original_source)
+    indent_match = match('^(\s+)', original_source[0])
+    if indent_match:
+        text_to_format = dedent(text_to_format)
+
     style_config = args.style
     if style_config is None and not args.no_local_style:
       style_config = file_resources.GetDefaultStyleForDir(os.getcwd())
     reformatted_source, changed = yapf_api.FormatCode(
-        py3compat.unicode('\n'.join(original_source) + '\n'),
+        py3compat.unicode(text_to_format),
         filename='<stdin>',
         style_config=style_config,
         lines=lines,
         verify=args.verify)
-    sys.stdout.write(reformatted_source)
+    out = reformatted_source
+    if indent_match:
+        out = indent(out, indent_match.group(0))
+    sys.stdout.write(out)
     return 2 if changed else 0
 
   files = file_resources.GetCommandLineFiles(args.files, args.recursive,
@@ -210,13 +227,13 @@ def FormatFiles(filenames,
           print_diff=print_diff,
           verify=verify,
           logger=logging.warning)
-      if has_change and reformatted_code is not None:
-        file_resources.WriteReformattedCode(filename, reformatted_code,
-                                            in_place, encoding)
       changed |= has_change
     except SyntaxError as e:
       e.filename = filename
       raise
+    if reformatted_code is not None:
+      file_resources.WriteReformattedCode(filename, reformatted_code, in_place,
+                                          encoding)
   return changed
 
 
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -37,7 +37,7 @@ from yapf.yapflib import py3compat
 from yapf.yapflib import style
 from yapf.yapflib import yapf_api
 
-__version__ = '0.6.3'
+__version__ = '0.7.1'
 
 
 import sys
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -30,12 +30,6 @@ from __future__ import print_function
 import argparse
 import logging
 import os
-from re import match
-
-try:
-    from textwrap import dedent, indent
-except ImportError:
-    from py3compat import indent
 
 from yapf.yapflib import errors
 from yapf.yapflib import file_resources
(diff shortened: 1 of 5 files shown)


purpleP:
I signed it!

gwelymernans:
Thank you. Could you squash the patches together?

yapf/yapflib/py3compat.py
```diff
@@ -83,3 +83,7 @@ class ConfigParser(configparser.ConfigParser):
 
     def read_file(self, fp, source=None):
       self.readfp(fp, filename=source)
+
+
+def indent(text, indent):
+    return '\n'.join(indent + l for l in text.split('\n')[:-1])
```
gwelymernans: How about this instead?

```
('\n'+indent).join(text.split('\n')[:-1])
```

yapf/__init__.py
```diff
@@ -143,16 +150,26 @@ def main(argv):
...
-        py3compat.unicode('\n'.join(original_source) + '\n'),
+        py3compat.unicode(text_to_format),
         filename='<stdin>',
         style_config=style_config,
         lines=lines,
         verify=args.verify)
-    sys.stdout.write(reformatted_source)
+    out = reformatted_source
+    if indent_match:
+        out = indent(out, indent_match.group(0))
```
gwelymernans: Do you need a terminating `\n` here?

yapf/__init__.py
```diff
@@ -143,16 +150,26 @@ def main(argv):
         original_source.append(py3compat.raw_input())
       except EOFError:
         break
+
+    original_source.append('')
+    text_to_format = '\n'.join(original_source)
```
gwelymernans: You need a terminating `\n` I think.

yapf/yapflib/py3compat.py
gwelymernans: Also, why is this function in `py3compat`? That module is for Python 3 compatibility functions only.

yapf/__init__.py
```diff
@@ -30,16 +30,23 @@
 import argparse
 import logging
 import os
-import sys
+from re import match
+
+try:
+    from textwrap import dedent, indent
+except ImportError:
+    from py3compat import indent
```
gwelymernans: Please move this to `py3compat.py`.

yapf/__init__.py
```diff
@@ -30,16 +30,23 @@
 import argparse
 import logging
 import os
-import sys
+from re import match
+
+try:
+    from textwrap import dedent, indent
```
gwelymernans: I prefer not importing `dedent` from `textwrap`, but using `textwrap.dedent` when necessary.

gwelymernans:
@purpleP Please squash all of your changes into one commit so that it's easier to review. Thanks! :-)

purpleP:
My eyes hurt after this, so I'll do that later. However I see all changes like one commit and don't know how to squash it.
23 апр. 2016 г. 2:26 пользователь Bill Wendling <notifications@github.com> написал:@purpleP Please squash all of your changes into one commit so that it's easier to review. Thanks! :-)

gwelymernans:
Basically, you just have to do `git rebase -i` and then mark all but the first commit as "squash". It's really best when there are several commits for a pull.

See more here:

  https://help.github.com/articles/about-git-rebase/

yapf/__init__.py
```diff
@@ -143,11 +143,13 @@ def main(argv):
...
       except EOFError:
         break
+
+    original_source.append('')
     style_config = args.style
     if style_config is None and not args.no_local_style:
       style_config = file_resources.GetDefaultStyleForDir(os.getcwd())
     reformatted_source, changed = yapf_api.FormatCode(
-        py3compat.unicode('\n'.join(original_source) + '\n'),
+        py3compat.unicode('\n'.join(original_source)),
```
gwelymernans: I'm not sure why you added an empty string to `original_source` just to get the terminating newline instead of doing it here. Is it more efficient?

yapf/__init__.py
```diff
@@ -210,13 +212,13 @@ def FormatFiles(filenames,
...
-      if has_change and reformatted_code is not None:
-        file_resources.WriteReformattedCode(filename, reformatted_code,
-                                            in_place, encoding)
       changed |= has_change
     except SyntaxError as e:
       e.filename = filename
       raise
+    if reformatted_code is not None:
+      file_resources.WriteReformattedCode(filename, reformatted_code, in_place,
+                                          encoding)
```
gwelymernans: Why did you move this down here? It no longer has the "`has_change`" check.

yapftests/main_test.py
```diff
@@ -98,7 +98,6 @@ def testEchoBadInput(self):
     bad_syntax = "  a = 1\n"
     with patched_input(bad_syntax):
       with captured_output() as (out, err):
-        with self.assertRaisesRegexp(SyntaxError, "unexpected indent"):
           ret = yapf.main([])
```
gwelymernans: Please dedent the `ret = yapf.main([])` line two spaces.

yapftests/yapf_test.py
```diff
@@ -15,6 +15,7 @@
 """Tests for yapf.yapf."""
 
 import io
+from re import match
```
gwelymernans: I prefer to use `import re` and use `re.match(...)` instead of importing `match`.

yapf/yapflib/yapf_api.py
```diff
@@ -116,9 +121,21 @@ def FormatCode(unformatted_source,
...
+
+  fst_newline = unformatted_source.find('\n')
+
+  indent_match = re.match('^(\s+)', unformatted_source[:fst_newline])
+  st = style.CreateStyleFromConfig(style_config)
+  if indent_match:
+    unformatted_source = dedent(unformatted_source)
+    original_indent = indent_match.group(0)
+    original_len = len(original_indent)
+    offset = original_len - original_len % 4
```
gwelymernans: What is the `4` in this line?

purpleP: it's default number of spaces for one level indent in python.

yapftests/main_test.py
purpleP: Sure, I'll do that later today.

yapf/__init__.py
purpleP: It should be, because if you add newline after joining you're creating a new string and when joining you don't

yapf/__init__.py
purpleP: Probably I was just tired.

purpleP: What's more important is that it passes tests with that.

yapf/yapflib/yapf_api.py
gwelymernans: The number of spaces is adjustable. See the `INDENT_WIDTH` in `style.py`.

yapftests/yapf_test.py
purpleP: Unless you import a lot of things from module it's useless and just a waste of time and space in lines. So without a good reason I surely won't do that.

yapf/__init__.py
gwelymernans: I believe the situation this was meant to solve was not writing to a file if there were no changes. Doing so updated the timestamp. It doesn't look like this is needed, though. So go ahead and move it back up please. :-)

yapftests/yapf_test.py
gwelymernans: How about "This is my project and I want it that way, and it keeps the namespace cleaner"?

yapf/yapflib/yapf_api.py
purpleP: so this should be style['INDENT_WITH'] right?

purpleP:
@gwelymernans I've tried git rebase -i but I don't think it worked. Also the proposal about indenting that you've maid wasn't working.

yapftests/yapf_test.py
purpleP: How about you do that yourself then.

purpleP: Like you have so polluted namespace anyway... It doesn't do any good, admit it =) I mean specifically in this case.

gwelymernans:
@purpleP I'm not sure why there was a difficulty with rebasing, or even creating a new pull request. I'll close this out, because it's not really reviewable in its current form and also the CLA checks are now failing. :-(

purpleP:
@gwelymernans I can create a new pull request if this would help. What is CLA?

gwelymernans:
@purpleP Yeah, a new pull request would work best here. For the CLA, see the @googlebot comment above.

PR #258, opened 2016-05-25
Fix typo in field name

ppolewicz:
I noticed that `_ParenState` class has no `break_before_paremeter` field. I *guess* that the author meant to use `split_before_parameter`, but no tests detect the change, so perhaps this line can be deleted altogether? Please investigate thoroughly before merging.


gwelymernans:
Thanks!

PR #259, opened 2016-05-26
Performance optimization

ppolewicz:
This gives a total of 25% performance boost on `yapf` project tree and a 50% boost on `whylog` tree (yapf can process x% more lines per second than it used to), both on python 2.7 and 3.4.

Please review and comment, but don't merge yet - I want to optimize further. I tried the cPickle trick instead of deepcopy and it was fast, but not as fast as the list comprehension.

### DIFF
--- a/yapf/yapflib/format_decision_state.py
+++ b/yapf/yapflib/format_decision_state.py
@@ -84,7 +84,7 @@ class FormatDecisionState(object):
 
   def Clone(self):
     new = copy.copy(self)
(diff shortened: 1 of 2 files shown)


gwelymernans:
Interesting, I'm surprised that `deepcopy` takes so long (I found out the same thing last night), and that the list comprehension is shorter.

Thank you!

ppolewicz:
One of the things that deepcopy does, is to find cycles in the group of objects that are being copied. We don't need that here.

PR #272, opened 2016-06-15
configure flake8 to match your style

bukzor:
Apparently you can't tell flake8/pep8 to use two-space indents, so I had to disable three checks.


bukzor:
Apparently you can't tell flake8/pep8 to use two-space indents, so I had to disable three checks.

PR #277, opened 2016-06-16
missing coverage: each of these changes should have broken a test

bukzor:
This patch is informational only; it isn't meant for merge.

Each of these changes should be a breaking change, but all tests are still passing.

This indicates missing tests.


gwelymernans:
Thanks for the example. Coverage is an ongoing process.

Issue #280, opened 2016-06-29
Yapf doesn't get odd indent width


b5y:
Hi!

Say, I want to set this indent:

```python
from yapf.yapflib.yapf_api import style, FormatFile


def my_own_style():
    my_style = style.CreatePEP8Style()
    my_style['INDENT_WIDTH'] = 3
    return my_style

# Lets format any file (with another indent) with this style:
formatted_code = FormatFile("any_file.py", 
                                style_config=style.SetGlobalStyle(my_own_style()), 
                                in_place=False)[0]
# And write it somewhere:
with open('resulted_insertion.py', 'w') as fd:
        fd.write(formatted_code)
````

You will not believe, but resulted file will be with 4 indent width instead of 3. The same is for 1 indent width and so on.
Please correct me if I am wrong.
If it is bug, I am ready to fix it.

**Comment by gwelymernans** [comment]
This does sound like a bug. If you'd like to attack it, please feel free. :-)

**Comment by b5y** [comment]
@gwelymernans That's great! Please assign this task to me. I will fix it on this week.

**Comment by gwelymernans** [comment]
It's not allowing me to assign it to you. I think you're not on the list or something. :-(

**Comment by b5y** [comment]
Okay. Got it

**Comment by b5y** [comment]
Another idea come to my mind during browsing yapf code. Why continuation indent must be the same as indent? 

For instance, in Python we could implement the following code (and it will be validated):

```
for x in range(10):
  for y in range(5): # <- 2 indents
     print x, y # <- 3 indents
```

PR #294, opened 2016-08-19
Fixing #280 bug

b5y:
Fix TODO on style module and #280 bug.

### DIFF
--- a/yapf/yapflib/style.py
+++ b/yapf/yapflib/style.py
@@ -313,7 +313,12 @@ def CreateStyleFromConfig(style_config):
   Raises:
     StyleConfigError: if an unknown style option was encountered.
   """
+  styles = (CreatePEP8Style(), CreateGoogleStyle(),
+             CreateFacebookStyle(), CreateChromiumStyle())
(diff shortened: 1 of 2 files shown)


b5y:
I found strange behavior on [testFormatFileDiff](https://github.com/google/yapf/blob/master/yapftests/yapf_test.py#L228) test. I have never used unittest before (always pytest) and was surprised about this problem.

The problem is not in test, actually. It is in unittest work. When I run all tests, this test does not pass. But if I run only this test (I am using PyCharm and python 2.7.11) - it passes!

Please fix me if something wrong. I am waiting for CI curiously.

b5y:
I signed it!

yapf/yapflib/style.py
```diff
@@ -423,5 +432,5 @@ def _CreateStyleFromConfigParser(config):
 # TODO(eliben): For now we're preserving the global presence of a style dict.
```
b5y: It is not good solution to make non-constant variable global for whole project.
Hope my solution is appropriate in this situation.

PR #296, opened 2016-08-29
Support optional spaces for default and named assigns.

alenz33:


gwelymernans:
Thanks, @alenz33. Could you add some tests? (`yapftest/reformatter_test.py` is probably the place you want.) Also update `CHANGELOG` and `README.rst`.

alenz33:
@gwelymernans 
Is that the correct way to update the CHANGELOG file?

gwelymernans:
This is good. Thank you!

PR #317, opened 2016-10-06
Using concurrent.futures to format multiple files at the same time

campos-ddc:
In my local use formatting the entire project went from 20s to 13s (13s being the time it takes to format the slowest file).

Should be safe for Python 2.7 and 3, fingers crossed fro Travis results :)

### DIFF
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -113,6 +113,12 @@ def main(argv):
       action='store_true',
       help="don't search for local style definition")
   parser.add_argument('--verify', action='store_true', help=argparse.SUPPRESS)
+  parser.add_argument(
+      '-p',
+      '--parallel',
+      action='store_true',
+      help=('Run yapf in parallel when formatting multiple files. Requires '
+            'concurrent.futures in Python 2.X'))
 
   parser.add_argument('files', nargs='*')
   args = parser.parse_args(argv[1:])
@@ -178,7 +184,8 @@ def main(argv):
       no_local_style=args.no_local_style,
       in_place=args.in_place,
       print_diff=args.diff,
-      verify=args.verify)
+      verify=args.verify,
+      parallel=args.parallel)
   return 0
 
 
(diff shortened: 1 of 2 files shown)


gwelymernans:
Thanks for this. We try to put all 2-3 compatibility stuff in `py3compat.py`. Could you refactor this so that this code will make a single call and the "heavy lifting" will be done in `py3compat.py`?

campos-ddc:
Tell me what you think about that approach. It makes `__main__` much cleaner, but kinda hides the fact that sometimes things aren't actually running in multiple processes.

Another option would be just asking `py3compat` if we can run `concurrent.futures`, which would still leave `__main__` with an `if yes, run the multiprocess function, otherwise run single process`

Any other suggestions are welcome too.

yapf/__init__.py
```diff
@@ -204,51 +204,21 @@ def FormatFiles(filenames,
...
-
-def _FormatFilesMultiProcess(filenames,
-                             lines,
-                             style_config=None,
-                             no_local_style=False,
-                             in_place=False,
-                             print_diff=False,
-                             verify=True):
-  import concurrent.futures
+  from yapflib.py3compat import ProcessPoolExecutor, as_completed
```
gwelymernans: No need to do this. Just use `py3compat.ProcessPoolExecutor` and `py3compat.as_completed`.

yapf/__init__.py
```diff
@@ -204,51 +204,21 @@ def FormatFiles(filenames,
...
   import multiprocessing
```
gwelymernans: Please put this import at the top of the file.

campos-ddc:
@gwelymernans I'll work on that when I have some time (sorry for the delay)

campos-ddc:
@gwelymernans see if that option makes sense to you. Right now it is always available, but using it in Python 2 without `concurrent.futures` will lead to an import error.

Also rebased on top of current master to avoid conflicts.

gwelymernans:
Thanks! I'm sorry for the delay. Things have been busy. I'll get to this this week.

PR #320, opened 2016-10-08
make sure input files are '.py' ended

AmeenAli:
make sure input files are python files

### DIFF
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -168,7 +168,7 @@ def main(argv):
   if not files:
     raise errors.YapfError('Input filenames did not match any python files')
   for FILE in files:
-	if FILE.split('.')[-1].lower() != "py":
+	  if FILE.split('.')[-1].lower() != "py":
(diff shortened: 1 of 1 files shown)


gwelymernans:
Is there a use case where not haven't this check caused a problem?

AmeenAli:
i don't even know why there is a problem! @gwelymernans

gwelymernans:
?? I don't understand what you're saying. At the moment, there isn't an issue with reformatting files that use Python syntax but don't have a `.py` extension. I don't think this PR is necessary.

AmeenAli:
what about files that does not have python syntax? @gwelymernans

gwelymernans:
There are two scenarios:

1. The files are explicitly listed on the command-line:

    ```shell
    python -m yapf file1
    ```

    In this case, we will process the files no matter what extension they have.

2. Recursive mode. In this case, we process files with a `.py` extension.

So there's really no need to reject any files that don't have that extension...

Issue #292, opened 2016-08-16
PEP 448 is unsupported


haphut:
yapf 0.11.0 with CPython 3.5.2 chokes on:
```py
foo = {'a': 1}
bar = {'b': 2}
baz = {**foo, **bar}
```

vmesel:
@gwelymernans how can I help solving this issue?

gwelymernans:
@vmesel I think the only thing that needs to be done is the `lib2to3` library needs to be updated to support those constructs.

haphut:
I think this is the relevant bug: https://bugs.python.org/issue25969

PR #352, opened 2016-12-29
Add handling for unpacking with dict expressions (currently 3.6 only)

bastih:
Only works on Python 3.6, since lib2to3 in Python 3.5 doesn't
recognize the unpacking token.

Prior to this patch,
```{**x}```
would raise an error like this in Python 3.5: 
```lib2to3.pgen2.parse.ParseError: bad input: type=36, value='**', context=('   ', (1, 11))```.

This is a known issue.

In Python 3.6, this is apparently fixed, so that we can indeed support PEP448 cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - 2.7
   - 3.4
   - 3.5
+  - 3.6
   - nightly
 
 matrix:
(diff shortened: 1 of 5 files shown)


bastih:
This should provide a fix for #292 (at least for Python 3.6, until those fixes are backported).

gwelymernans:
Sorry for the delay. Thank you for this!

bastih:
Oh no worries! Thank you for merging!

PR #361, opened 2017-01-28
Suggestion: VIM command

teoric:
I do not use YAPF so often, so I do not want a mapping. But calling the function is unhandy, too. Therefore I created a VIM command `:YAPF` that can take a range (see comments in code for how to use it).


teoric:
I signed.  (Though you might consider whether there is a minimal contribution that cannot contain enough creativity to justify such an agreement.)

gwelymernans:
Thanks! Unfortunately, there's nothing we can do about the CLA. :-(

PR #364, opened 2017-02-03
Fix missing `EACH_DICT_ENTRY_ON_SEPARATE_LINE` config

kobras8:
It will cause `Unknown style option` after used in `.style.yapf`


kobras8:
I signed it!

gwelymernans:
Thanks! There were others that were missing, so I combined them with your patch.

PR #366, opened 2017-02-04
Fix unittest to work on Windows (python 2 and 3)

driax:
tmpefile.NamedTemporaryFile does not work as expected. It will create a file with O_TEMPORARY on Windows which can only be opened by the same process and using the same flag.

So I have provided a alternative in a new file yapftest/utils.py, named NamedTempFile. I have also moved some other test utilities into this file: "TempFileContents" and "stdout_redirectory", both are context managers, as is NamedTempFile.

Note also that I have created a new test, which dumps --style-help to a file and tries to format a simple example using the style file. This would catch the "bug" that you momentarily introduced in master when you added the new knobs EACH_DICT_ENTRY_ON_SEPARATE_LINE SPLIT_BEFORE_DICT_SET_GENERATOR.

I have also fixed 2 small bugs that would occur on Python 2 on windows during the test cases. Mostly related to how newline transformation works. The still exist another bug that is neither fixed nor tested for specifically that "input" is used to read stdin which discards newlines and thus we have no idea what newlines to output. A proper fix for that would involve a small bit of refactoring to remove small unnecessary code.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,6 +14,9 @@
 ### Fixed
 - Split before all entries in a dict/set or list maker when comma-terminated,
   even if there's only one entry.
+- Will now try to set O_BINARY mode on stdout under Windows and Python 2.
+- Avoid unneeded newline transformation when writing formatted code to
+  output on (affects only Python 2)
 
 ## [0.15.2] 2017-01-29
 ### Fixed
--- a/yapf/yapflib/file_resources.py
+++ b/yapf/yapflib/file_resources.py
@@ -93,7 +93,7 @@ def WriteReformattedCode(filename,
   """
   if in_place:
     with py3compat.open_with_encoding(
-        filename, mode='w', encoding=encoding) as fd:
+        filename, mode='w', encoding=encoding, newline='') as fd:
       fd.write(reformatted_code)
   else:
     py3compat.EncodeAndWriteToStdout(reformatted_code)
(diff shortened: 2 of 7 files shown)


gwelymernans:
Thank you for the patches! One question, why did you add the `u` prefix to all of the strings in the tests?

driax:
the new TempFileContents context manager takes unicode arguments on Python 2, because the file is opened using `io.open` instead of just `open`. The makes the code semantically more similar between python 2 and 3, however it also means that python 2 will yell if you don't feed the write() call unicode strings. So I simply changed all the literals.
Maybe not shortest change. But it felt more modernized to base the code around stricter "io" file objects on both versions.

driax:
I just realized that the two versions of NamedTempFile is practically identical. I better clean that up.

gwelymernans:
Okay, sounds good. One last thing before I merge this. Could you update the `CHANGELOG` file for each PR?

driax:
When you say "for each PR" what do you mean exactly. Most of the commits only affect the tests. Only 7eadc3e "Python 2 standard/file output fixes." changes anything important. Should I just push a new commit with CHANGELOG updated with fixes:
- Will now try to set O_BINARY mode on stdout under Windows and Python 2.
- Avoid unneeded newline transformation when writing formatted code to
  output on (affects only Python 2)

gwelymernans:
You had two commits listed here, but no worries. The comments you have here are okay. Please go ahead and add the updated `CHANGELOG` to this pull request. Thanks!

Issue #328, opened 2016-10-29
USE_TABS does not work for indentation


spaceone:
USE_TABS=True is ignored for continued indentation. Instead spaces are used.

My configuration:

[.style.yapf](https://github.com/google/yapf/files/559934/default.style.yapf.txt)
Reproduce:

```
$ yapf --version
yapf 0.13.2
printf "property_descriptions = {\n\t'name': admin.property(\n\t\tshort_description=_('Foobar'),\n\t\tlong_description='',\n\t\tsyntax=syntax.hostName,\n\t\tmultivalue=False,\n\t\tinclude_in_default_search=True,\n\t\toptions=[],\n\t\trequired=True,\n\t\tmay_change=False,\n\t\tidentifies=True\n\t),\n}\n" > /tmp/foo.py
yapf --style .style.yapf /tmp/foo.py
```

Also you cannot specify /dev/stdin or "-" as argument for the file.

spaceone:
A follow up error of this is that with CONTINUATION_INDENT_WIDTH=4 the following gets replaced into wrong reindention:
```
layout = [
    Tab(_('General'), _('Basic settings'), layout=[
        Group(_('Computer account'), layout=[
            ['name', 'description'],
            'inventoryNumber',
        ]),
    ]),
]
```

→ 
```
layout = [
    Tab(_('General'), _('Basic settings'), layout=[
        Group(_('Computer account'), layout=[
   ['name', 'description'],
   'inventoryNumber',
        ]),
    ]),
]
```

This makes this tool unusable for me.

eskhool:
+1, facing the same issue with the tool. Any suggestions on where to start with the fix?

gwelymernans:
I'll take a look at this, but if you look at the `whitespace_prefix` bit in the `format_token` module, it should help determine what's going on here with the tabs.

PR #373, opened 2017-02-21
Fix USE_TABS=True is ignored for continued indentation

CaselIT:
This PR fixes the bug that caused spaces to be used for continued indentation when USE_TABS=True.

At the moment the output is not optimal since it ends up adding more tab than needed, since it just replaces the continued indentation spaced with tabs, but at leas the code is executable. I could not find a way to reliably reduce the number of tab to be added without outputting invalid code.

Closes https://github.com/google/yapf/issues/328

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@
   right-hand side. Split the arguments instead.
 - We need to split a dictionary value if the first element is a comment anyway,
   so don't force the split here. It's forced elsewhere.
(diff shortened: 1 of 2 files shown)


CaselIT:
I guess the coveralls bot was very pleased by this pr 😆

CaselIT:
@gwelymernans When is the next release scheduled? This bug in the current version is causing us some inconvenience. 
Thanks

Issue #375, opened 2017-02-23
Complex dicts get their values split needlessly. Regression from 0.15.2 to 0.16.0


do3cc:
I provided a test case that I will add as a PR after I finished writing this bug report.
I've used the test case to pin down the commit that added that behavior: 0e9b32114dc20a3b35854af012980795a118b03f
It is probably an unintended side effect.
The problem does not occur with the chromium style.

do3cc:
Correction, the problem does also occur with chromium style

gwelymernans:
Change b363bc uses an interesting way to reformat list-comprehensions. I was considering using that for dictionaries. That might simplify how dictionaries are handled, if you'd like to take a stab at it.

PR #376, opened 2017-02-23
Add testcase for dict problem

do3cc:
This is the test case for #375 
One test case is failing, the other isn't.


do3cc:
#375 is still open, is there anything else I can do to help getting this issue fixed?

PR #384, opened 2017-03-26
Remove the utf-8 byte-order mark

greut:
Currently `yapf` fails on files containing the infamous BOM. Those two commits are an _attempt_ to test and fix that issue.

This was the easiest way I found to test it but it's kind of agressive. I'm keen on knowing a better way.

Cheers,


gwelymernans:
Sorry, this fell through the cracks.  Could you place this as a function in `py3comat.py`?

greut:
Reopened there: https://github.com/google/yapf/pull/589

PR #395, opened 2017-04-21
Changes to prevent yapf splitting before the first argument in compound statements.

joachimmetz:


gwelymernans:
Please add a testcase for this.

joachimmetz:
Ack, will do. Note this might take a day or so.

Issue #405, opened 2017-05-09
Style api for use as a module


mpacer:
Right now there is no documented way to create a Style object for use with the "for use as a module" apis (i.e., `FormatCode` and `FormatFile`). 

Given the current apis (at least in as far as they are documented), the only way to use a custom style is to write out a file on disk and to point to that file. That's suboptimal if you're trying to use it as part of a conversion pipeline that tries to keep configuration and computation in-memory. It's also suboptimal for debugging a call to FormatCode in an interactive session (e.g., in a Jupyter notebook). 

Alternatively, having a way to set individual "knobs" directly in the module level APIs (e.g., by passing in a dictionary with the knob names as the keys) would avoid this as well. 

cc @Carreau

mpacer:
The problem seems to be more a problem of documentation than it is one of *support* per se. 

For example, the following works:
```python
test_input = '''def sum(a, b):
   """Sum two numbers."""     
   temp = (b &b & a  &     a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a  &    a&b &b & a)
   return temp
'''
style_config = """{
COLUMN_LIMIT:40,
SPLIT_PENALTY_AFTER_OPENING_BRACKET=2000,
}"""
formatted_code = FormatCode(test_input,style_config=style_config)
print(formatted_code[0])
```
with output: 

    def sum(a, b):
        """Sum two numbers."""
        temp = (b & b & a & a & b & b & a &
                a & b & b & a & a & b & b &
                a & a & b & b & a & a & b &
                b & a & a & b & b & a & a &
                b & b & a & a & b & b & a &
                a & b & b & a & a & b & b &
                a & a & b & b & a & a & b &
                b & a & a & b & b & a & a &
                b & b & a & a & b & b & a &
                a & b & b & a)
        return temp

However, figuring out how to make that work took longer than would be needed with slightly improved docs. Honestly, if I hadn't had a hunch that this should have been possible… I wouldn't have dove into the source and found https://github.com/google/yapf/blob/master/yapf/yapflib/style.py#L381-L383:

```python
if style_config.startswith('{'):
  # Most likely a style specification from the command line.
[...]

PR #407, opened 2017-05-10
Add ability to specify style via dict with string values

mpacer:
This actually is a better solution than just documenting how to write a string properly in the docs for in memory use, since you can just use the dict directly (see #405).

Becuase I wanted to keep it the same as before using the config parser, that means that all values need to be passed in as a string, which seems reasonable enough.

### DIFF
--- a/yapf/yapflib/style.py
+++ b/yapf/yapflib/style.py
@@ -375,17 +375,26 @@ def CreateStyleFromConfig(style_config):
     if not def_style:
       return _style
     return _GLOBAL_STYLE_FACTORY()
-  style_factory = _STYLE_NAME_TO_FACTORY.get(style_config.lower())
-  if style_factory is not None:
-    return style_factory()
-  if style_config.startswith('{'):
-    # Most likely a style specification from the command line.
-    config = _CreateConfigParserFromConfigString(style_config)
-  else:
-    # Unknown config name: assume it's a file name then.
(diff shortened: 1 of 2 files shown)


yapf/yapflib/style.py
```diff
@@ -375,18 +375,29 @@ def GlobalStyles():
...
-  style_factory = _STYLE_NAME_TO_FACTORY.get(style_config.lower())
-  if style_factory is not None:
-    return style_factory()
-  if style_config.startswith('{'):
-    # Most likely a style specification from the command line.
-    config = _CreateConfigParserFromConfigString(style_config)
-  else:
-    # Unknown config name: assume it's a file name then.
-    config = _CreateConfigParserFromConfigFile(style_config)
+  if isinstance(style_config, dict):
```
Carreau: `dict` >[ `collection.abc.mapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) ? Maybe a bit overreaching though.

mpacer: I think that would include counter and since all the values need to be strings for this to work properly, that seems dangerous? But maybe that means `isinstance` is the wrong way to go for this, since it allows subclasses to also count… maybe I should make a check explicitly that all the values are strings (or only pass through pairs to the config object if the value is a string)?

mpacer: so now strings are not required, so if someone passed in a counter it would still work.

PR #418, opened 2017-06-15
Update PEP 8 style to match modern PEP 8

njsmith:
PEP 8 was updated last year to prefer that line splits come before binary operators, rather than after:

https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

https://github.com/python/peps/commit/c59c4376ad233a62ca4b3a6060c81368bd21e85b

Really there should be a way to do this for other binary operations... for example, even with this setting, YAPF wants to do:

```
     orig_extracted = (
-        extract_tb(orig.__traceback__)
-        + extract_tb(orig.exceptions[0].__traceback__)
-        + extract_tb(orig.exceptions[0].exceptions[1].__traceback__))
+        extract_tb(orig.__traceback__) +
+        extract_tb(orig.exceptions[0].__traceback__) +
+        extract_tb(orig.exceptions[0].exceptions[1].__traceback__))
```

which is exactly what the example in PEP 8 says not to do. But this patch brings YAPF's PEP 8 style closer to what PEP 8 says.


njsmith:
cla signed

njsmith:
It looks like YAPF's handling of trailing commas is wrong too: PEP 8 says that if you have a trailing comma, the only two good options are to put the final bracket on a new line, or else to write a single element tuple like `(None,)` with no trailing space ([ref](https://github.com/python/peps/commit/c78ccb3dcb52c15b5177f19fa92fa97750e11b70)).

I'm not entirely sure how to express this in YAPF terms -- maybe `SPACE_BETWEEN_ENDING_COMMA_AND_CLOSING_BRACKET` should be `False`, and `SPLIT_ARGUMENTS_WHEN_COMMA_TERMINATED` should be `True`? (That's the opposite of how they're set now.)

Issue #415, opened 2017-06-02
Formatting regression


sbarzowski:
My `style.yapf`:
```
[style]
based_on_style = pep8
DEDENT_CLOSING_BRACKETS = true
```

I had a code like this:
```
def a():
    class B:
        def c(self):
            self.fields['lorem'].label = (
                'lorem ipsum dolor sit amet blah blah x'
            )
```
and yapf seemed to like it, however after recent upgrade from 0.16.1 to 0.16.2 it started reformatting it to:
```
def a():
    class B:
        def c(self):
            self.fields['lorem'
                        ].label = ('lorem ipsum dolor sit amet blah blah x')
```

It seems like a terrible place for a newline. Any idea why it happens?

jouve:
It looks related to 77ca21d9

joelnb:
Agreed that it looks like that commit is the issue but I'm not familiar enough with the code to fix it. I'll get reading though because currently I am unable to use the latest YAPF due to this issue.

joelnb:
I made a PR but (as I said there) not convinced it is the correct approach so would be interested in the opinions of others affected by this issue.

cybojenix:
in #419, you mentioned that #436 fixed your issue here.
I'm still experiencing the same bug referenced here.

When changing the column limit to 120, YAPF begins splitting the square brackets.

expected:
```python
...
            translated_data['date_finance_required'] = date_to_rfc3339(
                parse_date(translated_data['date_finance_required'])
            )
...
```

output:
```python
...
            translated_data['date_finance_required'
                            ] = date_to_rfc3339(parse_date(translated_data['date_finance_required']))

...
```

style config:
```ini
[style]
based_on_style = pep8
DEDENT_CLOSING_BRACKETS = true
COLUMN_LIMIT = 120
```

cybojenix:
Quick update, I tried with YAPF 0.16.1, and I still experience the above issue

mikelambert:
@cybojenix , the fix is here: https://github.com/google/yapf/commit/8a3b71f86b5c95b19629a778428a613881b03ca0  . And the fix says it is only in master and 0.18. According to pypi https://pypi.python.org/pypi/yapf, 0.18 was released on Sept 18th. Curious why you're using an older version 0.16.1 released in March?

cybojenix:
@mikelambert, sorry I didn't mention. The first comment was with 0.18.0. I tried 0.16.1 afterwards since it was mentioned that 0.16.2 introduced the regression.

[...]

PR #396, opened 2017-04-22
Changes to coalesce brackets not to line split before closing bracket

joachimmetz:

PR #419, opened 2017-06-24
Fix a regression related to coalesced brackets when using DEDENT_CLOSING_BRACKETS

joelnb:
This fixes https://github.com/google/yapf/issues/415 where the final bracket of a function call was not being dendented even with `DEDENT_CLOSING_BRACKETS` set.

I think this PR is probably currently the wrong approach but I wanted to start a discussion about what the correct behaviour here is. Given that the documentation for `COALESCE_BRACKETS` states that it is "Only relevant when `DEDENT_CLOSING_BRACKETS` is set." I think that in all cases the code produced by setting `COALESCE_BRACKETS` but not `DEDENT_CLOSING_BRACKETS` should be the same as setting neither option - otherwise the documentation should be changed.

I'm interested what others think though & am happy to make changes to this pull request (and add tests) as required.

### DIFF
--- a/yapf/yapflib/format_decision_state.py
+++ b/yapf/yapflib/format_decision_state.py
@@ -503,7 +503,7 @@ class FormatDecisionState(object):
           0, self.stack[-1].indent - style.Get('CONTINUATION_INDENT_WIDTH'))
 
       split_before_closing_bracket = True
-      if style.Get('COALESCE_BRACKETS'):
+      if style.Get('COALESCE_BRACKETS') and not style.Get('DEDENT_CLOSING_BRACKETS'):
         split_before_closing_bracket = False
 
       self.stack[-1].split_before_closing_bracket = split_before_closing_bracket


joelnb:
The reason for the Travis failures is because I have not modified the test added in https://github.com/google/yapf/pull/396 to reflect the change. When there is some consensus on what the correct behaviour is I will update the test to reflect that.

joelnb:
Signed!

rdrey:
We're also using `dedent_closing_brackets` and `coalesce_brackets` together and would like to keep the old behaviour. I've reverted to `0.16.1 ` for now.

joelnb:
I also found that this can cause code which is syntactically incorrect to be generated. This made up example:

```
if __name__ == "__main__":
    with open('/tmp/test.txt') as f:
        with some_resource:
            with another_resource:
                account = {
                    "username":
                        'admin',
                    "password":
                        'admin',
                    "validUntil":
                        int(time() + (6 * 7 * 24 * 60 * 60))  # in 6 weeks time
                }
```

Becomes:

```
if __name__ == "__main__":
    with open('/tmp/test.txt') as f:
        with some_resource:
            with another_resource:
                account = {
                    "username":
                        'admin',
                    "password":
                        'admin',
                    "validUntil":
                        int(time() +
                            (6 * 7 * 24 * 60 * 60))  # in 6 weeks time}
```

Note that the closing `}` has become part of the comment, meaning this will be a SyntaxError.

I'm still unsure whether this fix is the correct one or whether the condition here should simply be removed. I'm happy to either fix the tests to work with this fix or work on another one but will wait to hear opinions on the matter.

gwelymernans:
Whoa! That's a bad bug. :-(

gwelymernans:
Hmm...Good point. If `COALESCE_BRACKETS` is used, then it's probably okay to assume that the other option should also be set. What was the test failure that you saw?

joelnb:
The test failure is just a test which was added in the PR which added this condition.

My preference would be to remove this condition & the test case which goes along with it because I feel like it doesn't match the documentation (as explained above). I guess before doing that it might be nice to hear from the author of the PR which added this about their use case & see if we can come to a conclusion which makes everyone happy?

Either way it would be nice to add a test case for the code which is being formatted with a syntax error here as that will prevent a similar regression in the future.

gwelymernans:
I just merged #436. Please see if that takes care of this issue.

joelnb:
Yes that fixes the issue. Thanks for getting this sorted out!

PR #427, opened 2017-07-21
Configurable spaces around binary operators

jirikuncar:
Adds configuration options to allow current SymPy style for spaces around binary operators.


jirikuncar:
I signed it!

PR #428, opened 2017-07-21
Make yapf put the correct number of blank lines at --lines boundaries.

kevincox:
This allows YAPF to have the correct line counts in most cases. Namely
lines can be deleted only if they are in the range, but if either side
of a range needs correcting of blank lines more lines can be inserted to
get to the correct number.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,10 @@
 # All notable changes to this project will be documented in this file.
 # This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.16.4] Unreleased
+### Changed
+- Adjust blank lines on formatting boundaries when using the `--lines` option.
+
 ## [0.16.3] 2017-07-13
 ### Changed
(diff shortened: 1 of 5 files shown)


gwelymernans:
Thanks! And nice cleanup of `_MarkLinesToFormat`! Could you update the `CHANGELOG` file as well?

kevincox:
I couldn't quite figure out the CHANGELOG protocol so let me know if that works.

CHANGELOG
```diff
@@ -2,6 +2,10 @@
 # All notable changes to this project will be documented in this file.
 # This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.16.4] Unreleased
+### Fixed
```
gwelymernans: Please use `Changed` here and change the text to something like:

```
Insert missing blank lines on formatting boundaries when using the `--lines` option.
```

kevincox:
Fixed.

Issue #433, opened 2017-07-31
Return error if files were changed and quiet argument


mattdavis90:
What are the thoughts on having the a quiet argument to the command line application and returning a non-zero value if the code requires changing.

This would then allow chaining in a CI environment; e.g. test code formatting prior to running tests and hence stop the pull request if formatting is bad.

Thanks

gwelymernans:
This is interesting. Actually, the code doesn't follow the docstring. We're supposed to return a non-zero value if things changed, but we don't. Your change fixes that, so that's a plus. I'll take a closer look at it.

Issue #325, opened 2016-10-17
Exit code with diff


retr0h:
Diff returns an exit code of 0 regardless. This doesn't seem ideal since CI scripts can no longer rely on a exit code. I would expect diff to return a non-0 if it were to make a change. I understand #228 introduced this change, and I agree a 0 should be returned when making a format change, but running a diff is an entirely different use case, and IMO we should error when the files would be changed.

retr0h:
I see the [exit code is customizable](https://github.com/google/yapf/commit/6ac26c9ad032d532d717ce3eab6d45f585a9792b).  For some reason I don't have that CLI option.

```
[jodewey:~] 2 % yapf --version
yapf 0.13.1
[jodewey:~] % yapf --help
usage: yapf [-h] [-v] [-d | -i] [-r | -l START-END] [-e PATTERN]
            [--style STYLE] [--style-help] [--no-local-style]
            [files [files ...]]

Formatter for Python code.

positional arguments:
  files

optional arguments:
  -h, --help            show this help message and exit
  -v, --version         show version number and exit
  -d, --diff            print the diff for the fixed source
  -i, --in-place        make changes to files in place
  -r, --recursive       run recursively over directories
  -l START-END, --lines START-END
                        range of lines to reformat, one-based
  -e PATTERN, --exclude PATTERN
                        patterns for files to exclude from formatting
  --style STYLE         specify formatting style: either a style name (for
                        example "pep8" or "google"), or the name of a file
                        with style settings. The default is pep8 unless a
                        .style.yapf or setup.cfg file located in one of the
                        parent directories of the source file (or current
                        directory for stdin)
  --style-help          show style settings and exit; this output can be saved
                        to .style.yapf to make your settings permanent
  --no-local-style      don't search for local style definition
```

gwelymernans:
I believe that was reverted.

retr0h:
Ah, yes, you're right.  The question still remains.  A 0 exit code on diff seems incorrect, when diff is likely consumed by scripts in most cases.

rgreinho:
I am facing the same issue. 

As a workaround, I created a wrapper for YAPF which behaves like you described:

```
#!/bin/bash

DIR=${1:-.}
LINES=$(yapf -d -r "${DIR}" | wc -l | tr -s ' ')
exit ${LINES}
```

nvgoldin:
:+1: - it would be much better to customize the exit code for my usage case, or at least exit with something different than 0. this was the behaviour before, no?

[...]

PR #434, opened 2017-08-01
Add quiet flag and return value if quiet

mattdavis90:
Adds a quiet flag to the command line that suppresses output and sets a return value. This could be useful for performing checks in CI prior to running tests.

Fixes Issue #433

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,11 +2,15 @@
 # All notable changes to this project will be documented in this file.
 # This project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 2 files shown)


mattdavis90:
I signed it!

mst-ableton:
@gwelymernans could you take a look at this PR? We'd like to use `yapf` in a CI pipeline, and having a check like this would be great.

mst-ableton:
On update it looks like https://github.com/google/yapf/issues/325 (mostly) solves this issue for us, it's just not documented anywhere :).

Issue #432, opened 2017-07-31
COALESCE_BRACKETS coalesces commas as well


AlecBenzer:
E.g.:

```shell
$ yapf << EOF       
[
1,
2,
3,
]
EOF
[
    1,
    2,
    3,
]
```
```shell
$ yapf --style="{coalesce_brackets: true}" << EOF
[
1,
2,
3,
]
EOF
[
    1,
    2,
    3, ]  # Bracket and comma put on same line.
```
```shell
$ yapf --style="{coalesce_brackets: true}" << EOF
[
1,
2,
3  # No comma.
]
EOF
[1, 2, 3]
```

Is this intentional? I wouldn't think a comma would be considered a 'bracket', and I'd expect that the first and second examples would produce the same output.

AlecBenzer:
Fixed by #436 

```shell
$ yapf --version
yapf 0.17.0                                                                                                                                                                             
$ yapf --style="{coalesce_brackets: true}" << EOF
[
1,
2,
3,
]
EOF
[
    1,
    2,
    3,
]
```

PR #436, opened 2017-08-03
Make COALESCE_BRACKETS and DEDENT_CLOSING_BRACKETS work together

wallrj:
Attempt to address #432 

The tests for COALESCE_BRACKETS seem wrong to me: 
 * the example code in `yapftests/reformatter_basic_test.py` only has single brackets, with a new line and  dedented closing bracket. And the expected output demonstrates the removal of the newline, rather than coalesced brackets.
 * and while the integration tests in `yapftests/yapf_test.py` do demonstrate `COALESCE_BRACKETS`, the use of facebook base style, hides the bug, where `COALESCE_BRACKETS` overrides the `DEDENT_CLOSING_BRACKETS` behaviour.

I'm tempted to write more tests for the combination of those two options, or to add a argparse assertion to raise an error when `DEDENT_CLOSING_BRACKETS=False` is combined with `COALESCE_BRACKETS=True`; since that the README implies that `COALESCE_BRACKETS` does not apply in that case:

```
``COALESCE_BRACKETS``
    Do not split consecutive brackets. Only relevant when
    ``DEDENT_CLOSING_BRACKETS`` is set. For example:
```

### DIFF
--- a/yapf/yapflib/format_decision_state.py
+++ b/yapf/yapflib/format_decision_state.py
@@ -502,11 +502,7 @@ class FormatDecisionState(object):
       self.stack[-1].closing_scope_indent = max(
           0, self.stack[-1].indent - style.Get('CONTINUATION_INDENT_WIDTH'))
 
-      split_before_closing_bracket = True
-      if style.Get('COALESCE_BRACKETS'):
-        split_before_closing_bracket = False
-
-      self.stack[-1].split_before_closing_bracket = split_before_closing_bracket
+      self.stack[-1].split_before_closing_bracket = True
 
     # Calculate the split penalty.
     penalty = current.split_penalty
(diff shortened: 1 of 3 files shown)


wallrj:
I signed it!

joelnb:
This fixes the same issue I was trying to get rid of in https://github.com/google/yapf/pull/419. It's quite serious given that it can cause syntax errors. I would support this PR over the one I made because I feel like this fits best with the documented behaviour in the README.

gwelymernans:
Thank you for the fix!

AlecBenzer:
Fixes #432:

```shell
$ yapf --version
yapf 0.17.0                                                                                                                                                                             
$ yapf --style="{coalesce_brackets: true}" << EOF
[
1,
2,
3,
]
EOF
[
    1,
    2,
    3,
]
```

Thanks!

PR #437, opened 2017-08-03
fix FormatFile return value in print mode

adimascio:
API expects `yapf_api.FormatFile` to return a couple `(diff, changed)`
where `changed` must be a boolean.

If `changed` is not a boolean, it then breaks on:
```
  changed |= _FormatFile(...)
```


adimascio:
I signed it!

PR #438, opened 2017-08-03
Fix #204 and fix #339

### LINKED ISSUE #204: mistaken added indentation to comment when previous block ended in a comment
The specific issue is the final change, adding a level of indentation of the comment "attached" to the return statement, presumably because the previous block ended with a comment.  While this is flipping where the comment was attached (eg: end of block, then before block), I think that it's worth preserving the original location of that final comment because of the explicit newline, outdent, code at outdent pattern.

``` diff
         except Exception as ex:
             self.logger.exception(
                 "Caught exception while fetching livehoststatus. "
-                "Falling back to old API: %s",
-                ex
+                "Falling back to old API: %s", ex
             )
             # @todo dpittman 2015-09-25: I feel like this is way too broad a
             # catch statement for the failure mode assumption, but it matches
             # the original.  We should probably fix that.

-        # ...guess we try the old fashioned way.
+            # ...guess we try the old fashioned way.
         return self.get_bad_hosts_with_live_maint_status(maint, stage)

     def get_bad_hosts_with_live_host_status(self, maint, stage):
```

### LINKED ISSUE #339: yapf produces invalid syntax when dealing with inconsistent comment indentation
When dealing with inconsistent comment indentation, yapf (0.13.2) produces invalid syntax. This occurs around try/except blocks.

```
> yapf --version
yapf 0.13.2
```

**Example**
```python
try:
    print('hello')
    # test 1
# test 2
except Exception:
    pass
```
yapf produces:
```python
try:
    print('hello')
    # test 1
# test 2
:
except Exception
    pass
```

bouweandela:
Improve the handling of subsequent comments at decreasing levels of indentation.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,8 @@
 ## [0.17.1] UNRELEASED
 ### Fixed
 - Allow semicolons if the line is disabled.
+- Fix issue where subsequent comments at decreasing levels of indentation
+  were improperly aligned and/or caused output with invalid syntax.
 
 ## [0.17.0] 2017-08-20
 ### Added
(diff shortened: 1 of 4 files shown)


gwelymernans:
Please add testcases and update the `CHANGELOG` file.

gwelymernans:
Thanks!

Issue #315, opened 2016-10-05
Option to `SPLIT_ARGUMENTS_WHEN_COMMA_TERMINATED` even when there is only one argument


campos-ddc:
Right now we have this current format behavior:

```python
foo(a=1, b=2,)
foo(a=1,)

# formats to:

foo(
    a=1,
    b=2,
)
foo(a=1,)
```

i.e., `foo(a=1,)` stays as is because there is only one comma in that line. I'd like an option to indent that, making the final behavior:

```python
foo(a=1, b=2,)
foo(a=1,)

# formats to:

foo(
    a=1,
    b=2,
)
foo(
    a=1,
)
```

Current behavior seems to be intentional and hardcoded into `pytree_unrwapper._DetermineMustSplitAnnotation`

gwelymernans:
It will need to be a separate flag, something like `SPLIT_ALL_ARGUMENTS_WHEN_COMMA_TERMINATED`.

campos-ddc:
Sounds OK to me.

I guess that `pytree_unwrapper` is too internal to access `style` options right? If you agree, I'd need some tips on how to make a PR for this. If it is safe to access `style` there, I think I can handle it.

lwm:
Potentially also solved by https://github.com/google/yapf/pull/550 (without comma issue).

PR #441, opened 2017-08-07
Attempt to fix #359, comma-term tuples

### LINKED ISSUE #359: How to prevent YAPF from joining lines in dictionaries and lists?
YAPF converts
```
DJANGO_TEMPLATES_OPTIONS = {
  "context_processors": []
}
```
to
```
DJANGO_TEMPLATES_OPTIONS = {"context_processors": []}
```
and I cannot find the option to disable such behavior.

Is it possible?

jjhuang:
I was running into the problem of having tuple values on individual lines when comma-terminated, as described in https://github.com/google/yapf/issues/359.

Basically, this returns to the original methodology of not breaking until there are 2+ items only for the case where the parent leaf starts with the parenthesis.

Let me know if this is a bad fix -- I thought this was a simple way to reconcile the opinions in that issue and it seemed to pass the unit tests. I based this on https://github.com/google/yapf/commit/94962e597d0d318d83fccb2490c8dd5fb27a235d, which was the commit that changed this behavior originally.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@
 - Corrected how `DEDENT_CLOSING_BRACKETS` and `COALESCE_BRACKETS` interacted.
 - Fix return value to return a boolean.
 - Correct vim plugin not to clobber edited code if yapf returns an error.
+- Ensured comma-terminated tuples with multiple elements are split onto separate lines.
 
 ## [0.16.3] 2017-07-13
 ### Changed
--- a/yapf/yapflib/pytree_unwrapper.py
+++ b/yapf/yapflib/pytree_unwrapper.py
@@ -335,7 +335,7 @@ def _AdjustSplitPenalty(uwline):
 def _DetermineMustSplitAnnotation(node):
   """Enforce a split in the list if the list ends with a comma."""
   if not _ContainsComments(node):
-    token = node.parent.leaves().next()
+    token = next(node.parent.leaves())
     if token.value == '(':
       if sum(1 for ch in node.children
              if pytree_utils.NodeName(ch) == 'COMMA') < 2:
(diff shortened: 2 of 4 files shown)


gwelymernans:
Thanks for your PR. Could you update the `CHANGELOG` file please? I'll review this some more this weekend.

jjhuang:
Any feedback here?

gwelymernans:
Apologies for the late response. Merged. Thanks!

LEW21:
Before this change, YAPF worked perfectly for me:
```
[style]
based_on_style = pep8
column_limit = 9999
dedent_closing_brackets = true
```

When comma-terminated, all function call arguments were put onto their own lines - even when there is only a single argument.

```
func(
    a=5,
)
```

After the change, YAPF compresses this into a single line:
```
func(a=5, )
```

I'd prefer to return to the old behavior, as with the new one trailing commas in argument lists are somewhat useless.

jjhuang:
@LEW21 Hm I mainly did this because I employ single item tuples (representing a const list) and it was really annoying to have a single item collection get broken out into three lines. Unfortunately the only way you can produce a single item tuple is by adding the ending comma even when I intentionally do not want to break lines -- so I had no way around that. (I don't usually have functions with a single argument and expect to continuously add more).

Both use cases seem valid to me, so this may be a matter of introducing another knob that specifically deals with the tuple corner case?

LEW21:
@jjhuang While in case of tuples not-splitting might be a good idea (even tho I'd still prefer a knob to let me split them), this doesn't really affect function calls. Isn't there a way to detect if these parentheses are for a tuple or for a function call?

BTW One of the affected callables is "dict", and I'm frequently using this syntax to declare dicts (as there is no need to quote field names).

gwelymernans:
One fix may be to just use the existing knob. Thoughts?

```diff
@@ -332,7 +333,8 @@ def _DetermineMustSplitAnnotation(node):
   """Enforce a split in the list if the list ends with a comma."""
   if not _ContainsComments(node):
     token = next(node.parent.leaves())
-    if token.value == '(':
+    if (token.value == '(' and
+        not style.Get('SPLIT_ARGUMENTS_WHEN_COMMA_TERMINATED')):
       if sum(1 for ch in node.children
              if pytree_utils.NodeName(ch) == 'COMMA') < 2:
         return
```

LEW21:
This was already suggested in https://github.com/google/yapf/issues/315 - and is a good idea; but it probably should not affect tuples, only arguments.

PR #442, opened 2017-08-08
Add verbose flag to print out filenames as they are processed

vmax:
I didn't see a way to print out filenames as they are processed, so I decided to add `-vv` flag

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -24,6 +24,7 @@
 ### Changed
 - Adjust blank lines on formatting boundaries when using the `--lines` option.
 - Return 1 if a diff changed the code. This is in line with how GNU diff acts.
+- Add `-vv` flag to print out file names as they are processed
 ### Fixed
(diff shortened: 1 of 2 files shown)


vmax:
@googlebot I signed it!

gwelymernans:
There's already a `logging.info` message there. Perhaps we should use that instead of a `print` statement?

vmax:
@gwelymernans Is there a way to set logging level? Environment variable or something? (or should I introduce it instead as command line param)?

gwelymernans:
From the looks of it, there needs to be a command line flag and have it set at the beginning. That might be a bit overkill for what we need. For your fix, I think we can just get rid of the `logging` line right before it and keep the `print` message you set here. Could you also update the `CHANGELOG` file?

vmax:
@gwelymernans updated as you requested in `4da6322`, please have a look

gwelymernans:
Thank you!

Issue #458, opened 2017-09-30
Function definition formatting change in 0.17 -> 0.18


njsmith:
In 0.17, with [this .style.yapf](https://github.com/python-trio/trio/blob/4d41a6ddfa352cc2002592a64b58f9a7172c9495/.style.yapf), this code is formatted like this:

```python
async def open_file(
    file,
    mode='r',
    buffering=-1,
    encoding=None,
    errors=None,
    newline=None,
    closefd=True,
    opener=None
):
    pass

async def run_sync_in_worker_thread(
    sync_fn, *args, cancellable=False, limiter=None
):
    pass
```

If I keep everything else the same and upgrade to 0.18, then it gets reformatted as:

```python
async def open_file(file, mode='r', buffering=-1, encoding=None, errors=None,
                    newline=None, closefd=True, opener=None):
    pass


async def run_sync_in_worker_thread(sync_fn, *args, cancellable=False,
                                    limiter=None):
    pass
```

It's not really obvious to me which is better, but I was caught by surprise. Is this expected? I don't see anything about it in the CHANGELOG...

njsmith:
Bizarrely, `git bisect` says that the change happened in 6412ce704a8fcbf74544d12f2ee211d9b2087f9b.

Before that commit, there is a weird discrepancy between `def` and `async def`, where for some reason the argument list above becomes indented *more* (8 spaces) if I used `def`, but otherwise they're formatted similarly. After that commit, we get the effect shown above.

gwelymernans:
Ironically, it looks like the `split_before_first_argument` knob is causing this.

PR #459, opened 2017-09-30
Treat 'async def' like 'def'

njsmith:
yapf has special treatment for (a) compound statements, and (b)
compound statements that are not 'def' statements. So you would expect
that 'async def' would be in category (a), but not in category (b).

Originally, 'async def' was not in either category, but this was fixed
in 6412ce704a8f, which made 'async def' (and other async constructs)
treated like compound statements. However, the special check for 'def'
statements was *not* extended to 'async def', so now 'async def' was
in categories (a) *and* (b).

This commit fixes that, so that 'async def' is in category (a) but not
category (b).

Fixes gh-458

### DIFF
--- a/yapf/yapflib/format_decision_state.py
+++ b/yapf/yapflib/format_decision_state.py
@@ -180,7 +180,7 @@ class FormatDecisionState(object):
     # Prevent splitting before the first argument in compound statements
     # with the exception of function declarations.
     if (style.Get('SPLIT_BEFORE_FIRST_ARGUMENT') and
-        self.line.first.value != 'def' and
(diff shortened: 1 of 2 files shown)


njsmith:
There are probably other places where `async def` and `def` are treated differently that this didn't fix -- e.g. `CanMergeMultipleLines` does a check for `lines[0].first.value in {'def', 'class'}`.

gwelymernans:
I'm sorry. I didn't see this PR before I submitted my change. It's similar to your change.

njsmith:
Wow, eerily similar indeed :-)

PR #470, opened 2017-10-29
Improve formatting for comprehensions and generators.

msuozzo:
Terminology reference:
```
[loop_var for loop_var in a_list if loop_var]
 ^      ^ ^                    ^ ^         ^
 \-expr-/ \------comp_for------/ \-comp_if-/
```

Implements the following preferences:
- Try to keep comprehensions on a single line.
- Prefer splitting at comp_if when comp_for was split as well.
- Try to avoid splitting at comp_for when expr is exactly one token long
  - This keeps the (often short) clauses together: [a for a in ...].

Also: Fixes a bug in the penalty logic where comprehensions provided as
arguments to functions had a heavy penalty for splitting at their start.

Previous Behavior:

```python
obj.FunctionCall(loop_var.Get()
                 for loop_var in a_very_long_list_name)
```

Fixed Example:

```python
obj.FunctionCall(
    loop_var.Get() for loop_var in a_very_long_list_name)
```


msuozzo:
I've tweaked the implementation to more reliably favor single-line comprehensions.

yapf/yapflib/pytree_unwrapper.py
```diff
@@ -333,7 +333,8 @@ def _DetermineMustSplitAnnotation(node):
   if not _ContainsComments(node):
     token = next(node.parent.leaves())
     if token.value == '(':
-      if sum(1 for ch in node.children
+      if sum(1
```
gwelymernans: This change has be a little concerned. I know it's a corner case, but I find this new version to be less readable. Perhaps you could not split when the variable is a single integer or name? What are your thoughts?

yapf/yapflib/style.py
```diff
@@ -178,6 +178,9 @@ def method():
       after."""),
     SPLIT_BEFORE_NAMED_ASSIGNS=textwrap.dedent("""\
       Split named assignments onto individual lines."""),
+    SPLIT_COMPLEX_COMPREHENSIONS=textwrap.dedent("""\
+      Set to True to prefer splitting list comprehensions and generators
+      when they have non-trivial expressions and/or filter clauses."""),
```
gwelymernans: Please add an entry for this knob in `README.rst`.

CHANGELOG
```diff
@@ -10,6 +10,9 @@
   dictionary that doesn't fit on a single line.
 - Improve splitting of elements in a tuple. We want to split if there's a
   function call in the tuple that doesn't fit on the line.
+- Improve splitting of comprehensions and generators. Prefer to keep
+  comprehensions on a single line but if a line-break is required, prefer to
+  split complex expressions and 'if' clauses onto their own lines.
```
gwelymernans: Please add this to an `Added` section:

```
## [0.19.1] UNRELEASED
### Added
- Improve splitting of comprehensions and generators. Prefer to keep
  comprehensions on a single line but if a line-break is required, prefer to
  split complex expressions and 'if' clauses onto their own lines.
```

yapf/yapflib/pytree_unwrapper.py
msuozzo: Yeah I was thinking the same thing. I think the one-token expression is a good idea. Now if only Python allowed `len(generator)`...

msuozzo:
Updated to address comments.

gwelymernans:
It looks like the CI is failing. Some of the test cases need to be updated it looks like.

gwelymernans:
Thanks!

PR #501, opened 2018-01-09
Split vim plugin file into two scripts

xoox:
Have you ever tested the plugin under Vim 8?
An error of `E492: Not an editor command: YAPF` will arise when you type in `:YAPF`.
It seems that commands can't be defined in autoload script in Vim 8. This PR will fix the command issue in Vim 8.
FYI. If this PR is going to be merged, I've signed the CLA.


gwelymernans:
THanks!

PR #518, opened 2018-01-30
Support python2 unicode strings in CreateStyleFromConfig

### LINKED ISSUE #517: CreateStyleFromConfig crashes from any Unicode string in Python 2.7
When using the `yapflib` YAPF API under Python 2.7, passing a Unicode string to `style.CreateStyleFromConfig` causes YAPF to crash out like this:

```
  def testDefaultBasedOnExplicitlyUnicodeTypeString(self):
    cfg = style.CreateStyleFromConfig(u'{}')
    self.assertIsInstance(cfg, dict)
```
...
```
  File "/Users/brasmith/projects/yapf/yapftests/style_test.py", line 258, in testDefaultBasedOnExplicitlyUnicodeTypeString
    cfg = style.CreateStyleFromConfig(u'{}')
  File "/Users/brasmith/projects/yapf/yapf/yapflib/style.py", line 451, in CreateStyleFromConfig
    return _CreateStyleFromConfigParser(config)
UnboundLocalError: local variable 'config' referenced before assignment
```
(The test case above is a new one I'm adding.)

The cause of the problem here is that `style.CreateStyleFromConfig` is doing an `isinstance` check _only_ for the `str` type. That's fine for Python 3, but it only covers old-style byte strings in Python 2. When a Unicode string is passed into this function, none of the conditions that _should_ create the `config` variable are hit, and it ultimately raises the aforementioned `UnboundLocalError`.

infinitewarp:
Closes #517


infinitewarp:
Also, I'm aware that I don't have the CLA yet. I'm working with (read: "waiting on") my employer to get that cleared…

infinitewarp:
Are there any comments or objections to merging this? @gwelymernans or other maintainers?

Issue #513, opened 2018-01-26
Extra spaces being added by yapf (ad-infinitum).


alanhdu:
Using `yapf 0.20.1` and Python 3.6.4, when I `yapf` the following code:

```python
import yaml

{
   "config": {
           **yaml.load(f"""
    a:
      b: 0""")},
 }
```

yapf adds an extraneous 13 spaces after the closing `"""`:

```
--- test.py	(original)
+++ test.py	(reformatted)
@@ -4,5 +4,5 @@
    "config": {
            **yaml.load(f"""
     a:
-      b: 0""")},
+      b: 0"""             )},
  }
```

Running `yapf` repeatedly just adds even more spaces (ad-infinitum). This does *not* happen if:
- We don't use a format string
- We take out the wrapping `config` dictionary
- We don't use the new dict unpacking `**` syntax.

This also ignores `yapf: disable` comments, so 
```python
import yaml
# yapf: disable
{
   "config": {
           **yaml.load(f"""
    a:
      b: 0""")},
 }
# yapf: enable
```

*still* inserts extra spaces.

gwelymernans:
Whoa! Yeah, this is bad...

gwelymernans:
I'm not able to replicate this. Could you try with HEAD?

alanhdu:
After `pip install git+https://github.com/google/yapf@master` I still get the same behavior.

Hm... could you try using the following dockerfile then (where `test.py` is the code sample at the beginning?). That might take care of any environment mismatches.

```dockerfile
FROM python:3.6

RUN pip install yapf
ADD test.py /test.py
RUN yapf -d /test.py
```

gwelymernans:
I don't have Docker...:-(

alanhdu:
Hm... what system are you using? I'm on Ubuntu 16.04 and Python 3.6.4 (installed w/ Anaconda) -- are you on a different platform?

Alternatively, I'm also happy to contribute a patch (although it might take a while for me to get to it) -- do you have some intuition about where in the code base I should start poking at?

gwelymernans:
I tried it on both Linux (Debian install) and Mac and it didn't replicate. If you'd like to do a patch, I would appreciate it. The code to look at is `format_token::AddWhitespacePrefix()`. Whatever is adding the spaces to the code does it through that function.

alanhdu:
Huh -- I think I know where the reproducibility problem is.

[...]

PR #519, opened 2018-01-30
Expand list of allowable string prefixes

alanhdu:
Closes https://github.com/google/yapf/issues/513

### DIFF
--- a/yapftests/format_token_test.py
+++ b/yapftests/format_token_test.py
@@ -19,6 +19,7 @@ from lib2to3 import pytree
 from lib2to3.pgen2 import token
 
 from yapf.yapflib import format_token
+from yapf.yapflib import py3compat
 
 
 class FormatTokenTest(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


alanhdu:
I'm not a fan of just listing all the allowable prefixes in the Python 3 case, but my regex-fu is weak and I'm wasn't sure of a better way to do it.

gwelymernans:
There might be a more compact way to write it, but that's how it's specified in the Python docs, so let's just keep it the way you have it.

Could you update the `CHANGELOG` file and add a test please?

alanhdu:
@gwelymernans I've added a unit test. Is that good enough, or do you prefer to also add an integration / regression test?

yapftests/format_token_test.py
```diff
@@ -32,6 +33,16 @@ def testSimple(self):
     self.assertEqual("FormatToken(name=COMMENT, value=# A comment)", str(tok))
     self.assertTrue(tok.is_comment)
 
+  def testIsMultilineString(self):
```
gwelymernans: This should go into `reformatter_python3_test.py`. It has support for PY3 stuff.

Issue #527, opened 2018-02-14
add a knob to modify the number of blank lines around top level function and class definitions


Anthchirp:
[PEP8 says use 2](https://www.python.org/dev/peps/pep-0008/#blank-lines). As things go a different number is of course much more sensible, so make it something adjustable.

This forms part of #347.

PR #528, opened 2018-02-14
Add knob BLANK_LINES_AROUND_TOP_LEVEL_DEFINITION

Anthchirp:
suggested patch for #527 to gauge whether there is any interest in this at all.

If so, need to
- [ ] make sure it actually works
- [ ] add test
- [ ] discuss if there is a nicer way of expressing 1+number_of_lines
- [ ] ...or should it be a "1 or 2" boolean flag instead of an int?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,8 @@
 - Introduce a new option of formatting multiline literals. Add
   `SPLIT_BEFORE_CLOSING_BRACKET` knob to control whether closing bracket should
   get their own line.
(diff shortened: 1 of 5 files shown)


gwelymernans:
@Anthchirp Thanks for the PR. Could you sign the CLA for this? :-)

Anthchirp:
signed

gwelymernans:
Could you update the `CHANGELOG` and `README.rst` files with the new flag?

Issue #380, opened 2017-03-09
use_tabs=True does not work for lists and dictionaries


eskhool:
We use_tabs for indentation but  continuation_indent_width still uses spaces instead of tabs for lists and dictionaries that are split over multiple lines. If required there should be a separate knob for this which allows setting of tabs vs spaces for lists and dictionaries

``

gwelymernans:
I'm not a "tabs" guy, so I need to ask more about what you need. Should the tabs in the continuation indent be the same number as the value of the `CONTINUATION_INDENT` knob? So if it's 4, then there would be 4 tabs?

eskhool:
Thanks for not starting a 'holy war' on this 👍 . Ideally, though this may vary even amongst us 'tabs' people, I think I would want the continuation indent to be additional to the indent at which the code is at right now. so CONTINUATION_INDENT=1 should result in the code below
Eg below
`
def a():
	call_to_a_long_function_(
		b=2,
		c=3,
	)

`

gwelymernans:
I'm not above holy wars, but we need to support the feature. ;-)

I think I understand what you want then. I'm going to create a branch and let you test it out for me, if you would.

eskhool:
Us minority of 'tabbers' definitely need the support of the larger community :D.
Happy to test, with a reasonably large and varied code base.

yota-code:
I agree with @eskhool: an opening character (brace, bracket, etc.) give me an extra indent (ie. tab) on the next line.
As a "tabber" too, I can not understand the trend to do ASCII-art and space based alignment :), where you break everything each time you change a name... 

Nonetheless, thank you for yapf

eskhool:
@yota-code, as a fellow tabber ...are you able to use yapf at all without this change? I find that it completely breaks my formatting. Just clarifying if that thank you if for the effort in general or because it works for you also :)

LEW21:
+1 about just using extra normal indent, as a rule there is simply no place in the tab world for visual indents.

@eskhool I'm not.

LEW21:
Actually, it seems that it works correctly on yapf 0.16.1+.

PetterS:
I think using tabs only for indentation is also a common case. For example, this is supported by clang-format as "UseTab: ForIndentation". Then spaces can still be used to vertically align stuff.

gregwym:
Seems like it will be part of v0.20.2. Very exciting to see this lands!

gregwym:
[...]

PR #531, opened 2018-02-27
Add `CONTINUATION_ALIGNMENT_TYPE` knob to adjust how continuation lines vertically aligned when `USE_TABS` = True

yinyin:
The following alignment modes are implemented:

- `SPACE`: vertically align with spaces. current (0.20.2) and default behavior.
- `FIXED`: just use fixed number of tabs (`CONTINUATION_INDENT_WIDTH` / `INDENT_WIDTH`).
- `LESS`: Slightly left if cannot vertically align continuation lines with tabs.
- `MORE`: Slightly right if cannot vertically align continuation lines with tabs.

Here is how different mode look like:

continuation_alignment_type

with configuration:

```YAML
[style]
based_on_style = chromium
USE_TABS = true
COLUMN_LIMIT=32
INDENT_WIDTH=4
CONTINUATION_INDENT_WIDTH=8
# CONTINUATION_ALIGNMENT_TYPE =
```

Test cases are added. But the function `_TabModeContAlignPadding()` I added to `format_token.py` for generating alignment padding string still marked as *miss* in coverage report. Tried to fix but no clue so far.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,8 @@
 - Introduce a new option of formatting multiline literals. Add
   `SPLIT_BEFORE_CLOSING_BRACKET` knob to control whether closing bracket should
   get their own line.
+- Added `CONTINUATION_ALIGNMENT_TYPE` knob to choose vertical alignment style
+  when `USE_TABS` is enabled.
 
 ## [0.20.2] 2018-02-12
 ### Changed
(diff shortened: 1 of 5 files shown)


yinyin:
I signed it!

yinyin:
As this patch is related to #380 , I should post there first to see if this approach is usable for others. Sorry for the noise.

PR #542, opened 2018-03-22
Knob to enforce blank line between module docstring and modeline/shebang

### LINKED ISSUE #208: Space between modeline/shebang and module docstring
Imagine this file:

``` python
#!/usr/bin/env python
# -*- coding: utf-8 name> -*-

"""Some module docstring."""


def foobar():
    pass
```

Running yapf on it produces this diff:

``` diff
--- yapf.py (original)
+++ yapf.py (reformatted)
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 name> -*-
-
 """Some module docstring."""

```

I would actually have expected yapf **not** to remove the blank line between the shebang/encoding and the module docstring (i.e. do nothing).

Stealthii:
This PR adds functionality to allow the placement of a newline between the module's docstring, and the modeline/shebang if it exists.

Fixes #208.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,8 @@
 - Introduce a new option of formatting multiline literals. Add
   `SPLIT_BEFORE_CLOSING_BRACKET` knob to control whether closing bracket should
   get their own line.
+- Add 'BLANK_LINES_AROUND_TOP_LEVEL_DEFINITION' knob to control the number
+  of blank lines between top-level function and class definitions.
 - Added `CONTINUATION_ALIGN_STYLE` knob to choose continuation alignment style
   when `USE_TABS` is enabled.
 ### Fixed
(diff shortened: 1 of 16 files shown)


Stealthii:
I signed it!

Stealthii:
As recommended I've kept the behaviour the same by default for all code styles - if someone wants to review this and check that it shouldn't be different for sub-styles, that would be great.

In addition, this knob only allows change of enforcement, the following is not possible:
* Leaving the spacing as is, rather than enforcing style (although I believe this is outside yapf's scope)
* Defining how many lines (if any) should be left between modeline/shebang and the module's docstring (I personally do not see this being a requirement - a future PR by an affected party could implement if needed)

Stealthii:
Interested to know what happened with the merge commit here? I'm pretty sure I had a single commit that had `google/yapf/master` as the parent - it looks like 32ac922 was merging a copy of the 'master' branch that had diverged from upstream (rebased maybe)

gwelymernans:
I had to do a manual merge to be able to merge this PR. Was the resulting merge incorrect?

PR #548, opened 2018-03-31
Bug: Does not exclude hidden folders in a relative context

DHager:
This bug arises when the root directory is specified as a relative path, and when you are trying to exclude an immediate child directory (or file) which starts with a period. The exclusion pattern is fine, but the file path(s) it is being compared against are each over-stripped so that the period is no longer present.

In this example, the exclusion filter has failed to prevent changing `dependency.py`

    user@host:/var/project# file ./.serverless
    ./.serverless: directory

    user@host:/var/project# yapf -vv --recursive --exclude '*/.serverless/*' -i .
    Reformatting ./.serverless/requirements/dependency.py
    [--snip--]

For anyone discovering this PR who has the same problem, a workaround is to use an absolute path to the current directory instead:

    user@host:/var/project# yapf -vv --recursive --exclude '*/.serverless/*' -i `pwd`
    Reformatting /var/project/acme/exceptions.py
    [--snip--]

### DIFF
--- a/yapf/yapflib/file_resources.py
+++ b/yapf/yapflib/file_resources.py
@@ -148,7 +148,10 @@ def _FindPythonFiles(filenames, recursive, exclude):
 
 def IsIgnored(path, exclude):
   """Return True if filename matches any patterns in exclude."""
-  return any(fnmatch.fnmatch(path.lstrip('./'), e.rstrip('/')) for e in exclude)
+  path = path.lstrip("/")
+  while path.startswith("./"):
+    path = path[2:]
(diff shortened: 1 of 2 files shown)


DHager:
Hmm, I'm guessing the coverage/coveralls check failure is spurious, since it refers to "uncovered" lines that have no apparent relationship to my commits...

DHager:
Hi guys, is there anything I can add or document to help move this issue along?

gwelymernans:
Apologies about the delay. Thank you for the fix!

PR #550, opened 2018-04-04
Added an option to split on all args even if not comma terminated.

CardenB:
I noticed that there was an option to split args to individual lines, but only when comma terminated. I wanted an option to do this on all args, so I implemented something that seemed to work.

What else is needed to get this change accepted? I reviewed the contributing guidelines and seem to be in compliance with all of them, but there were not many.

### DIFF
--- a/yapf/yapflib/format_decision_state.py
+++ b/yapf/yapflib/format_decision_state.py
@@ -375,6 +375,13 @@ class FormatDecisionState(object):
         if previous.value in '(,':
           if opening.matching_bracket.previous_token.value == ',':
             return True
+    if style.Get('SPLIT_ARGUMENTS'):
+      # Split before arguments in a function call or definition if the
+      # arguments are terminated by a comma.
+      opening = _GetOpeningBracket(current)
+      if opening and opening.previous_token and opening.previous_token.is_name:
+        if previous.value in '(,' and opening.matching_bracket:
+          return True
 
     if ((current.is_name or current.value in {'*', '**'}) and
         previous.value == ','):
--- a/yapf/yapflib/style.py
+++ b/yapf/yapflib/style.py
@@ -175,6 +175,8 @@ _STYLE_HELP = dict(
     SPLIT_ARGUMENTS_WHEN_COMMA_TERMINATED=textwrap.dedent("""\
       Split before arguments if the argument list is terminated by a
       comma."""),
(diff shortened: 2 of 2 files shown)


CardenB:
This should fix #issue461

CardenB:
Will be adding unit tests to increase code coverage

pelzhaus:
Any updates here? Looking forward to this one

lwm:
Yes @CardenB! Awesome, I really want to see this land soon.

I think something like this might work for the test:

```
  def testSplittingAllArgs(self):
      style.SetGlobalStyle(style.CreateStyleFromConfig('{split_arguments: true}'))
      unformatted_code = textwrap.dedent("""\
          responseDict = {"timestamp": timestamp, "someValue":   value, "whatever": 120}
          """)
      expected_formatted_code = textwrap.dedent("""\
          responseDict = {
              "timestamp": timestamp,
              "someValue":   value,
              "whatever": 120
          }
          """)
      uwlines = yapf_test_helper.ParseAndUnwrap(unformatted_code)
      self.assertCodeEqual(expected_formatted_code, reformatter.Reformat(uwlines))
```

I was looking in `yapftests/reformatter_basic_test.py`.

CardenB:
@lwm thanks for the tests! I'm looking back into this now. It didn't work as expected before, and I got distracted.

CardenB:
@pelzhaus @lwm want to take another look? It passes tests but I'm paranoid that I'm breaking something, because the logic seems too simple.

CardenB:
Change seems good. Realized CI was failing because of code formatting issues lol. @gwelymernans what else can I do to improve this PR?

gwelymernans:
Apologies about they wait. Please add the new flag to the README file as well.

CardenB:
@gwelymernans How can I go about creating a new release so that people are able to take advantage of this without installing from source?

I saw some instructions in https://github.com/google/yapf/blob/master/HACKING.rst which are helpful. I think I can follow them, but I'm betting I'll run into some permissions issues, especially when pushing to pypi or something. Is this something that you or some other admin usually do? or should I do this myself?

lwm:
Huzzah, thanks @CardenB, @gwelymernans!

pelzhaus:
🍾 thanks all! @CardenB @gwelymernans @lwm

Issue #544, opened 2018-03-26
YAPF inserts weird long breaks in function call


alanhdu:
Using both `yapf 0.21` and the current master (`8d90c204e8abecc59d8e9a6037782293f26fa7f8`), `yapf` formats the following function call:

```python
f(
    long_function_name(16, g(lrotate) % 16),
    long_function_name(16, g(rrotate) % 16))
```

into the much uglier:

```python
f(
    long_function_name(16,
                       g(lrotate) % 16), long_function_name(
                           16,
                           g(rrotate) % 16))
```

If we use a longer function name then `g`, then `yapf` formats

```python
f(
    long_function_name(16, func(lrotate) % 16),
    long_function_name(16, func(rrotate) % 16))
```

into
```python
f(
    long_function_name(16,
                       func(lrotate) % 16),
    long_function_name(16,
                       func(rrotate) % 16))
```

I'm not sure why the line breaks are necessary (the `long_function_name` calls fit comfortably under the 80 character line limit).

mnsaab:
@alanhdu Are you able to recreate this issue with multiple other cases? If so, would you please be able to attach an image with one or two of those cases.

alanhdu:
@mnsaab Sorry, but I don't think I understand the request fully -- should I come up with other reproducible test cases (do the ones I provided not reproduce?) or should I literally just screenshot my text editor with the examples I provided?

jbrill:
We were able to reproduce the issue. Working on a solution now.

gwelymernans:
That is definitely ugly!! And your second example is unusual given each argument should be on its own line (as you mentioned).

@jbrill The way I would suggest attacking this issue would be to look into increasing the split penalty for each argument in a function. There is a lot of logic in `format_decision_state.py::MustSplit()` is used to enforce a lot of splitting decisions. However, some of the logic is getting "unwieldy". I would like to move away from `MustSplit()` and relying more on split penalties to drive the formatting.

gwelymernans:
IMO, having the arguments to `f()` on individual lines makes sense:

```python
f(
    long_function_name(16, func(lrotate) % 16),
    long_function_name(16, func(rrotate) % 16))
```

jbrill:
Was able to resolve the issue by scrapping some of the logic around line 400 regarding is_func_call. Specifically, changing:
`
if (not self._FitsOnLine(current, opening.matching_bracket) 
or ((opening.matching_bracket.next_token and 
[...]

PR #557, opened 2018-04-15
"YAPF inserts weird long breaks in function call" fix

jbrill:
This fix required simply a change in conditional logic regarding MustSplit() in format_decision_state.py

I've added two unit tests as well in reformatter_basic_test.py to encompass the two example tests provided by @alanhdu in issue #544


mnsaab:
@gwelymernans Hey Bill, Jason and I both have the CLAs signed for our respective emails connected to these github accounts. However, the Googlebot keeps saying that we failed. Any ideas why?

gwelymernans:
Weird. I checked with the database and it showed that you have signed it. It must just be an oversight.

gwelymernans:
Oh yeah, you will want to run YAPF on the YAPF codebase (that's why the travis-ci check is failing).

```shell
python -m yapf -r -i .
```

jbrill:
Thanks! Seems to still be failing when SCA=True environmental variables are set. I’ll definitely dig deeper into my own changes as well to encompass the original logic.

-Jason

keis:
I'm running into the same issue trying out yapf, are you still working on this @jbrill?

PR #558, opened 2018-04-16
Add documentation about --diff return code

mst-ableton:
Didn't see any documentation on here about the return code, and hoped to save time for other people looking to accomplish the same thing. Let me know if there's a better place in the doc to put this.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -126,6 +126,17 @@ Options::
       -vv, --verbose        Print out file names while processing
 
 
+------------
+Return Codes
+------------
+
+Normally YAPF returns zero on successful program termination and non-zero otherwise.
+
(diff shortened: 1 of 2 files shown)


gwelymernans:
Thanks! Though maybe instead of making this one of the FAQs, we should have something further up that indicates what the return values are in different situations and what they mean. Maybe in the `Usage` section? What do you think?

mst-ableton:
Sure, that sounds reasonable. I'll make an edit in a bit.

mst-ableton:
@gwelymernans ok, let me know if this is more to your liking!

gwelymernans:
Thank you!

PR #570, opened 2018-05-20
initial version, include missing files in MANIFEST.in

meeuw:
LICENSE file is missing from tarball, which is recommended to include for the Fedora package.

I've also added more missing files.


meeuw:
I've signed the CLA.

PR #589, opened 2018-07-23
py3compat: remove any BOM

greut:
This is a redo of #384 as it got lost...

----

Currently yapf fails on files containing the infamous BOM. Those two commits are an attempt to test and fix that issue.

This was the easiest way I found to test it but it's kind of agressive. I'm keen on knowing a better way.

Cheers,

### DIFF
--- a/yapf/yapflib/py3compat.py
+++ b/yapf/yapflib/py3compat.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Utilities for Python2 / Python3 compatibility."""
 
+import codecs
 import io
 import os
 import sys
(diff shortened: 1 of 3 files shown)


yapf/yapflib/py3compat.py
```diff
@@ -116,3 +117,12 @@ class ConfigParser(configparser.ConfigParser):
 
     def read_file(self, fp, source=None):
       self.readfp(fp, filename=source)
+
+def removeBOM(source):
```
gwelymernans: Please run `yapf` over the code first. (The lines should be two-space indented.)

greut:
:+1: thanks @gwelymernans !

PR #592, opened 2018-08-01
Adds an INDENT_BLANK_LINES option.

hugo-ricateau:
Adds an INDENT_BLANK_LINES option in order to allow indentation to continue on blank lines. Might be useful when trying to copy and paste pieces of code directly into a Python interpreter.

### DIFF
--- a/yapftests/reformatter_basic_test.py
+++ b/yapftests/reformatter_basic_test.py
@@ -133,6 +133,40 @@ class BasicReformatterTest(yapf_test_helper.YAPFTest):
     uwlines = yapf_test_helper.ParseAndUnwrap(unformatted_code)
     self.assertCodeEqual(expected_formatted_code, reformatter.Reformat(uwlines))
 
+  def testIndentBlankLines(self):
+    try:
+      style.SetGlobalStyle(
+          style.CreateStyleFromConfig(
+              '{based_on_style: chromium, indent_blank_lines: true}'))
+      unformatted_code = textwrap.dedent("""\
+          class foo(object):
+
+            def foobar(self):
+
+              pass
+
+            def barfoo(self, x, y):  # bar
+
+              if x:
+
+                return y
+
+
+          def bar():
+
+            return 0
+          """)
+      expected_formatted_code = """class foo(object):\n  \n  def foobar(self):\n    \n    pass\n  \n  def barfoo(self, x, y):  # bar\n    \n    if x:\n      \n      return y\n\n\ndef bar():\n  \n  return 0\n"""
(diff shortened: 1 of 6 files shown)


yapftests/reformatter_basic_test.py
```diff
@@ -133,6 +133,40 @@ def foobar():  # foo
...
+              if x:
+
+                return y
+
+
+          def bar():
+
+            return 0
+          """)
+      expected_formatted_code = """class foo(object):\n  \n  def foobar(self):\n    \n    pass\n  \n  def barfoo(self, x, y):  # bar\n    \n    if x:\n      \n      return y\n\n\ndef bar():\n  \n  return 0\n"""
```
gwelymernans: Is there a reason to put the expected formatting string on one line?

hugo-ricateau: I definitely agree that this one lined code is pretty ugly...

However, despite it might be difficult to read,
```
      expected_formatted_code = """class foo(object):\n  \n  def foobar(self):\n    \n    pass\n  \n  def barfoo(self, x, y):  # bar\n    \n    if x:\n      \n      return y\n\n\ndef bar():\n  \n  return 0\n"""
```
has two advantages over
```
      expected_formatted_code = """class foo(object):
  
  def foobar(self):
    
    pass
  
  def barfoo(self, x, y):  # bar
    
    if x:
      
      return y


def bar():
  
  return 0
"""
```
First, it highlights that some lines contain only white spaces. Secondly, it avoids to have an indented string with a different level than its context.

A third option might be to write it this way:
```
      expected_formatted_code = ("""class foo(object):\n"""
                                 """  \n"""
                                 """  def foobar(self):\n"""
                                 """    \n"""
                                 """    pass\n"""
                                 """  \n"""
                                 """  def barfoo(self, x, y):  # bar\n"""
                                 """    \n"""
                                 """    if x:\n"""
                                 """      \n"""
                                 """      return y\n"""
                                 """\n"""
                                 """\n"""
                                 """def bar():\n"""
                                 """  \n"""
                                 """  return 0\n""")
```
Which is probably a good compromise between the two other propositions.

Let me know your preference and I'll update the code.

hugo-ricateau-tiime: @gwelymernans ready for review.

hugo-ricateau: @gwelymernans ready for review.

gwelymernans:
Sorry for the late reply. Could you please add an entry to the `CHANGELOG` file?

PR #618, opened 2018-09-16
Catch KeyboardInterrupt when reading from stdin.

MichaelGrupp:
Do not print traceback and directly exit with error code.

This is more consistent with other UNIX CLI tools.


MichaelGrupp:
CLA signed.

PR #637, opened 2018-11-02
Use canonical name for snake case in contributing docs

dschaller:


dschaller:
@gwelymernans mind taking a look at this?

Issue #541, opened 2018-03-20
Align trailing comment?


0017031:
Is it possible to align the trailing comments?
Like the "AlignTrailingComments" option from clang-format.

gwelymernans:
YAPF doesn't support that, because it's against most style guides. I'd be happy to take a PR adding it behind a knob. :-)

PR #649, opened 2018-11-19
Updated trailing comment horizontal alignment based on a list of potential values

davidbrownell:
This addresses https://github.com/google/yapf/issues/541.

Thanks!

### DIFF
--- a/yapf/yapflib/style.py
+++ b/yapf/yapflib/style.py
@@ -181,7 +181,39 @@ _STYLE_HELP = dict(
       before each trailing comment) or list of of values (representing
       alignment column values; trailing comments within a block will
       be aligned to the first column value that is greater than the maximum
-      line length within the block)."""),
+      line length within the block). For example:
+      
+      With spaces_before_comment=5:
+
+        1 + 1 # Adding values
+
+      will be formatted as:
+
+        1 + 1     # Adding values <-- 5 spaces between the end of the statement and comment
+
+      With spaces_before_comment=15, 20:
+
+        1 + 1 # Adding values
+        two + two # More adding
+
+        longer_statement # This is a longer statement
+        short # This is a shorter statement
+
+        a_very_long_statement_that_extends_beyond_the_final_column # Comment
+        short # This is a shorter statement
+
+      will be formatted as:
+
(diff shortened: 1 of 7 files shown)


gwelymernans:
Could you squash these commits? It's hard to review this way.

yapf/yapflib/style.py
```diff
@@ -176,7 +176,12 @@ def method():
     SPACES_AROUND_DEFAULT_OR_NAMED_ASSIGN=textwrap.dedent("""\
       Use spaces around default or named assigns."""),
     SPACES_BEFORE_COMMENT=textwrap.dedent("""\
```
gwelymernans: This should come with an example of how to use it. (Similarly for the README.)

yapftests/yapf_test.py
```diff
@@ -1436,5 +1436,247 @@ def testSimple(self):
     self._Check(unformatted_code, expected_formatted_code)
 
 
+class HorizontallyAlignedTrailingCommentsTest(unittest.TestCase):
```
gwelymernans: Please add some examples with disabled lines.

davidbrownell: Unfortunately, I'm not sure what you mean by disabled lines. Do you mean tests where spaces_before_comment is set to an integer rather than a list?

gwelymernans: Lines which are within a disabled section. Either they have a `# yapf: disable` comment on the line or around the line, or they're in a section not covered by the `--lines` option.

So something like:

```
# yapf: disable
func()  # foo
b() # b
# yapf: enable
```

None of those comments should be modified. Similarly:

```
func()  # yapf: disable
b() # a comment
g() # align with b()'s comment
```

In this `func()`'s comment shouldn't be aligned, but `b()`'s and `g()`'s should. (At least if I understand what you're going for with this feature. :-)

davidbrownell: Ah, got it. You are right in that this change as it is now will not respect this behavior. I'll add some tests and implement the functionality.

Thanks!

gwelymernans:
Could you update `CHANGELOG` please?

davidbrownell:
PR https://github.com/google/yapf/pull/653 includes the CHANGELOG updates.

PR #655, opened 2018-12-05
Updated `pre-commit` shell script

mmphego:
- Updated shebang from `#!/bin/bash` to `#!/usr/bin/env bash` for [portability](https://en.wikipedia.org/wiki/Shebang_%28Unix%29#Portability) related issues.
- Double quote to prevent globbing and word splitting.
- Using `command -v` instead of `which`: The reason to select the former over the latter is that it avoids a dependency on something that is outside your shell.

Signed-off-by: Mpho Mphego <mpho112@gmail.com>

### DIFF
--- a/plugins/pre-commit.sh
+++ b/plugins/pre-commit.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Git pre-commit hook to check staged Python files for formatting issues with
 # yapf.
@@ -21,15 +21,14 @@
 # is used.
 
 # Find all staged Python files, and exit early if there aren't any.
-PYTHON_FILES=(`git diff --name-only --cached --diff-filter=AM | \
-  grep --color=never '.py$'`)
-if [ ! "$PYTHON_FILES" ]; then
+PYTHON_FILES=$(git diff --name-only --cached --diff-filter=AM | grep --color=never '.py$')
+if [ -z "${PYTHON_FILES}" ]; then
   exit 0
 fi
 
 ########## PIP VERSION #############
 # Verify that yapf is installed; if not, warn and exit.
-if [ -z $(which yapf) ]; then
+if ! command -v yapf >/dev/null; then
   echo 'yapf not on path; can not format. Please install yapf:'
   echo '    pip install yapf'
   exit 2
@@ -37,7 +36,7 @@ fi
 ######### END PIP VERSION ##########
 
 ########## PIPENV VERSION ##########
-# if [ -z $(pipenv run which yapf) ]; then
+# if ! pipenv run yapf --version 2>/dev/null 2>&1; then
 #   echo 'yapf not on path; can not format. Please install yapf:'
 #   echo '    pipenv install yapf'
 #   exit 2
@@ -46,16 +45,12 @@ fi
 
 
 # Check for unstaged changes to files in the index.
-CHANGED_FILES=(`git diff --name-only ${PYTHON_FILES[@]}`)
-if [ "$CHANGED_FILES" ]; then
+CHANGED_FILES=$(git diff --name-only "${PYTHON_FILES[@]}")
+if [ -n "${CHANGED_FILES}" ]; then
   echo 'You have unstaged changes to some files in your commit; skipping '
   echo 'auto-format. Please stage, stash, or revert these changes. You may '
   echo 'find `git stash -k` helpful here.'
-  echo
-  echo 'Files with unstaged changes:'
-  for file in ${CHANGED_FILES[@]}; do
-    echo "  $file"
-  done
+  echo 'Files with unstaged changes:' "${CHANGED_FILES[@]}"
   exit 1
 fi
 
@@ -64,22 +59,18 @@ fi
 echo 'Formatting staged Python files . . .'
 
 ########## PIP VERSION #############
-yapf -i -r ${PYTHON_FILES[@]}
+yapf -i -r "${PYTHON_FILES[@]}"
 ######### END PIP VERSION ##########
 
 ########## PIPENV VERSION ##########
-# pipenv run yapf -i -r ${PYTHON_FILES[@]}
+# pipenv run yapf -i -r "${PYTHON_FILES[@]}"
 ###### END PIPENV VERSION ##########
 
 
-CHANGED_FILES=(`git diff --name-only ${PYTHON_FILES[@]}`)
-if [ "$CHANGED_FILES" ]; then
+CHANGED_FILES=$(git diff --name-only "${PYTHON_FILES[@]}")
+if [ -n "${CHANGED_FILES}" ]; then
   echo 'Reformatted staged files. Please review and stage the changes.'
-  echo
-  echo 'Files updated:'
-  for file in ${CHANGED_FILES[@]}; do
-    echo "  $file"
-  done
+  echo 'Files updated: ' "${CHANGED_FILES[@]}"
   exit 1
 else
   exit 0


plugins/pre-commit.sh
```diff
@@ -46,14 +45,14 @@ fi
...
-CHANGED_FILES=(`git diff --name-only ${PYTHON_FILES[@]}`)
-if [ "$CHANGED_FILES" ]; then
-  echo 'You have unstaged changes to some files in your commit; skipping '
-  echo 'auto-format. Please stage, stash, or revert these changes. You may '
-  echo 'find `git stash -k` helpful here.'
+CHANGED_FILES=$(git diff --name-only "${PYTHON_FILES[@]}")
+if [ "${CHANGED_FILES}" ]; then
+  echo "You have unstaged changes to some files in your commit; skipping "
+  echo "auto-format. Please stage, stash, or revert these changes. You may "
+  echo "find $(git stash -k) helpful here."
```
gwelymernans: Does `$(git stash -k)` print or is it expanded?

plugins/pre-commit.sh
```diff
@@ -21,23 +21,22 @@
...
-if [ ! "$PYTHON_FILES" ]; then
+PYTHON_FILES=$(git diff --name-only --cached --diff-filter=AM | grep --color=never '.py$')
+if [ ! "${PYTHON_FILES}" ]; then
   exit 0
 fi
 
 ########## PIP VERSION #############
 # Verify that yapf is installed; if not, warn and exit.
-if [ -z $(which yapf) ]; then
-  echo 'yapf not on path; can not format. Please install yapf:'
```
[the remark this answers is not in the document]
mmphego: Perhaps, I should put more attention to detail.
Thanks for pointing that out.

plugins/pre-commit.sh
```diff
@@ -46,14 +45,14 @@ fi
...
+CHANGED_FILES=$(git diff --name-only "${PYTHON_FILES[@]}")
+if [ "${CHANGED_FILES}" ]; then
+  echo "You have unstaged changes to some files in your commit; skipping "
+  echo "auto-format. Please stage, stash, or revert these changes. You may "
+  echo "find $(git stash -k) helpful here."
   echo
   echo 'Files with unstaged changes:'
-  for file in ${CHANGED_FILES[@]}; do
+  for file in "${CHANGED_FILES[@]}"; do
     echo "  $file"
```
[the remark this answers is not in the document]
mmphego: Fixed on next commit

plugins/pre-commit.sh
```diff
@@ -21,23 +21,22 @@
...
+if ! command -v yapf >/dev/null; then
   echo 'yapf not on path; can not format. Please install yapf:'
   echo '    pip install yapf'
   exit 2
 fi
 ######### END PIP VERSION ##########
 
 ########## PIPENV VERSION ##########
-# if [ -z $(pipenv run which yapf) ]; then
+# if ! command -v pipenv run command -v yapf >/dev/null; then
```
vapier: i don't think this works.  `command` is a shell built in, so trying to run it through `pipenv` will fail.  i think the fact this relies upon `pipenv` means also relying upon `which` is fine.  or you could try:

```bash
if ! pipenv run yapf --version 2>/dev/null 2>&1; then
```

plugins/pre-commit.sh
```diff
@@ -46,15 +45,15 @@ fi
...
+if [ "${CHANGED_FILES}" ]; then
   echo 'You have unstaged changes to some files in your commit; skipping '
   echo 'auto-format. Please stage, stash, or revert these changes. You may '
   echo 'find `git stash -k` helpful here.'
   echo
   echo 'Files with unstaged changes:'
-  for file in ${CHANGED_FILES[@]}; do
-    echo "  $file"
+  for file in "${CHANGED_FILES[@]}"; do
+    echo '  ${file}'
```
vapier: using single quotes here breaks ... you need double quotes.

if you really wanted to rewrite this, you could do:

```bash
printf '  %s\n' "${CHANGED_FILES[@]}"
```

plugins/pre-commit.sh
```diff
@@ -64,21 +63,21 @@ fi
...
-if [ "$CHANGED_FILES" ]; then
+CHANGED_FILES=$(git diff --name-only "${PYTHON_FILES[@]}")
+if [ -n "${CHANGED_FILES}" ]; then
   echo 'Reformatted staged files. Please review and stage the changes.'
   echo
   echo 'Files updated:'
-  for file in ${CHANGED_FILES[@]}; do
-    echo "  $file"
+  for file in "${CHANGED_FILES[@]}"; do
+    echo '  ${file}'
```
vapier: same here

plugins/pre-commit.sh
```diff
@@ -46,15 +45,15 @@ fi
 
 
 # Check for unstaged changes to files in the index.
-CHANGED_FILES=(`git diff --name-only ${PYTHON_FILES[@]}`)
-if [ "$CHANGED_FILES" ]; then
+CHANGED_FILES=$(git diff --name-only "${PYTHON_FILES[@]}")
+if [ "${CHANGED_FILES}" ]; then
```
vapier: would be nice if this used `-n` explicitly in the string test

plugins/pre-commit.sh
```diff
@@ -46,15 +45,15 @@ fi
...
   echo 'You have unstaged changes to some files in your commit; skipping '
   echo 'auto-format. Please stage, stash, or revert these changes. You may '
   echo 'find `git stash -k` helpful here.'
   echo
   echo 'Files with unstaged changes:'
-  for file in ${CHANGED_FILES[@]}; do
-    echo "  $file"
+  for file in "${CHANGED_FILES[@]}"; do
```
vapier: you don't need the for loop anymore.  the single printf statement below is sufficient.

plugins/pre-commit.sh
```diff
@@ -64,21 +63,21 @@ fi
...
-CHANGED_FILES=(`git diff --name-only ${PYTHON_FILES[@]}`)
-if [ "$CHANGED_FILES" ]; then
+CHANGED_FILES=$(git diff --name-only "${PYTHON_FILES[@]}")
+if [ -n "${CHANGED_FILES}" ]; then
   echo 'Reformatted staged files. Please review and stage the changes.'
   echo
   echo 'Files updated:'
-  for file in ${CHANGED_FILES[@]}; do
-    echo "  $file"
+  for file in "${CHANGED_FILES[@]}"; do
```
vapier: same here -- no need for a for loop

plugins/pre-commit.sh
mmphego: Att to detail, I missed that 
Thanx!

mmphego:
Ready to merge - Updated script.

plugins/pre-commit.sh
```diff
@@ -29,38 +29,34 @@ fi
...
   exit 2
 fi
 ######### END PIP VERSION ##########
 
 ########## PIPENV VERSION ##########
 # if ! pipenv run yapf --version 2>/dev/null 2>&1; then
-#   echo 'yapf not on path; can not format. Please install yapf:'
-#   echo '    pipenv install yapf'
+#   printf 'yapf not on path; can not format. Please install yapf:\n' \
+#     '    pipenv install yapf'
```
vapier: same here -- should just leave it as `echo`

plugins/pre-commit.sh
```diff
@@ -29,38 +29,34 @@ fi
 ########## PIP VERSION #############
 # Verify that yapf is installed; if not, warn and exit.
 if ! command -v yapf >/dev/null; then
-  echo 'yapf not on path; can not format. Please install yapf:'
-  echo '    pip install yapf'
+  printf 'yapf not on path; can not format. Please install yapf:\n' \
```
vapier: this unfortunately prints a message that lacks a trailing newline which can be bad for a CLI tool

you should just leave this comment block alone as `echo` works fine here

plugins/pre-commit.sh
```diff
@@ -29,38 +29,34 @@ fi
...
-  echo 'You have unstaged changes to some files in your commit; skipping '
-  echo 'auto-format. Please stage, stash, or revert these changes. You may '
-  echo 'find `git stash -k` helpful here.'
-  echo
-  echo 'Files with unstaged changes:'
-  for file in "${CHANGED_FILES[@]}"; do
-    printf '  %s\n' "${CHANGED_FILES[@]}"
-  done
+if [ -n "${CHANGED_FILES}" ]; then
+  printf 'You have unstaged changes to some files in your commit; skipping ' \
```
vapier: this won't print correctly.  you should leave the `echo` statements alone and only convert the for loop to a single printf call.

plugins/pre-commit.sh
```diff
@@ -73,12 +69,8 @@ yapf -i -r "${PYTHON_FILES[@]}"
 
 CHANGED_FILES=$(git diff --name-only "${PYTHON_FILES[@]}")
 if [ -n "${CHANGED_FILES}" ]; then
-  echo 'Reformatted staged files. Please review and stage the changes.'
-  echo
-  echo 'Files updated:'
-  for file in "${CHANGED_FILES[@]}"; do
-    printf '  %s\n' "${CHANGED_FILES[@]}"
-  done
+  printf 'Reformatted staged files. Please review and stage the changes.\n'  \
```
vapier: same here

plugins/pre-commit.sh
```diff
@@ -29,38 +29,34 @@ fi
...
+    'auto-format. Please stage, stash, or revert these changes. You may ' \
+    'find `git stash -k` helpful here.\n' \
+    'Files with unstaged changes:  %s\n' "${CHANGED_FILES[@]}"
   exit 1
 fi
 
 # Format all staged files, then exit with an error code if any have uncommitted
 # changes.
-echo 'Formatting staged Python files . . .'
+printf 'Formatting staged Python files . . .\n'
```
vapier: this works, but isn't that useful.  i'd just leave the `echo` be.

plugins/pre-commit.sh
mmphego: reverted back to `echo`

mmphego:
Ready to merge - Updated script.
@vapier

mmphego:
> Could you squash all of the commits into one? It's hard to review them right now.

My apologies - I have since updated the PR.

gwelymernans:
Thanks!

Issue #654, opened 2018-12-03
yapf PR or fork?


davidbrownell:
I was thinking about adding a couple of features for yapf - they are fairly specific to my personal coding style so I'd understand if they would be more appropriate in a fork rather than yapf itself. Let me know what you think.

1) I tend to apply alignment formatting to assignment statements in model-, class-, and __init__-level statements...

The code: 

```
module_level_assignment = 10

class Foo(object):
    class_level_assignment = 20

    def __init__(self):
        self._a = 30
        self._b = 40
```

would become:

```
module_level_assignment                = 10

class Foo(object):
    class_level_assignment             = 20

    def __init__(self):
        self._a                        = 30
        self._b                        = 40

```

Internally, I'd refactor https://github.com/google/yapf/pull/649 and have that code and this code use similar logic. I'd probably also include functionality to specify the levels where alignment would be permitted (model-, class-, __init__-, any-level).

2) I add spaces around dict/list/tuple delimiters...

The code:

```
[1, 2, 3]
{'a': 'b'}
(x, y, z)
```

would become:

```
[ 1, 2, 3 ]
{ 'a' : 'b' }
( x, y, z ) # This one is tricky, as it will be interesting to differentiate between tuple delimiters and function calls, logical groupings, etc.
```

I'd put this functionality behind a switch that defaults to False (don't add spaces around delimiters).

What do you think - PR or fork?

Thanks!

PR #657, opened 2018-12-08
Added functionality to insert spaces around dicts, lists, and tuples.

davidbrownell:
Implemented part 2 of https://github.com/google/yapf/issues/654.

This change will insert spaces after opening- and before closing-tokens for dicts, lists, and tuples. There are 3 settings to control behavior for all of those types independently. The behavior will remain the same as it was before the change (no added spaces) if the settings are not provided.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,9 @@
 ### Added
 - `ALLOW_SPLIT_BEFORE_DEFAULT_OR_NAMED_ASSIGNS` allows us to split before
   default / named assignments.
+- Added `SPACES_AROUND_LIST_DELIMITERS`, `SPACES_AROUND_DICT_DELIMITERS`,
+  and `SPACES_AROUND_TUPLE_DELIMITERS` to add spaces after the opening-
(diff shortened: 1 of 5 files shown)


davidbrownell:
> 
> 
> [Coverage Status](https://coveralls.io/builds/20528169)
> 
> Coverage decreased (-0.03%) to 95.479% when pulling **[99d06d0](https://github.com/google/yapf/commit/99d06d045568360d0b60456e2507c13958153148) on davidbrownell:list_dict_tuple_spaces** into **[f672d7b](https://github.com/google/yapf/commit/f672d7b1073c29054cfcc9c32f344c948e8cf1bf) on google:master**.

It looks like the decreased code coverage numbers are based on a defense-in-depth-style check. My gut feeling is that these are good checks (I can image a scenario where some sort of code refactor might light up these scenarios) but they can't be hit based on the way that the code is written. 

However, I can change these statements to asserts to keep the coverage numbers up if you like.

Let me know if you have a preference.

PR #672, opened 2019-02-08
BUGFIX: Operator spacing

harahu:
Fixes a bug where applying binary operators to strings would
result in the blacklist specified in
`NO_SPACES_AROUND_SELECTED_BINARY_OPERATORS` to not take effect.

### DIFF
--- a/yapf/yapflib/unwrapped_line.py
+++ b/yapf/yapflib/unwrapped_line.py
@@ -291,9 +291,9 @@ def _SpaceRequiredBetween(left, right):
       # If there is a type hint, then we don't want to add a space between the
       # equal sign and the hint.
       return False
-    if rval not in '[)]}.':
+    if rval not in '[)]}.' and not right.is_binary_op:
       # A string followed by something other than a subscript, closing bracket,
-      # or dot should have a space after it.
+      # dot, or a binary op should have a space after it.
       return True
   if left.is_binary_op and lval != '**' and _IsUnaryOperator(right):
     # Space between the binary operator and the unary operator.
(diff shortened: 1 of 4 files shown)


yapf/yapflib/unwrapped_line.py
```diff
@@ -291,9 +291,9 @@ def _SpaceRequiredBetween(left, right):
       # If there is a type hint, then we don't want to add a space between the
       # equal sign and the hint.
       return False
-    if rval not in '[)]}.':
+    if rval not in '[)]}.' and not right.is_binary_op:
       # A string followed by something other than a subscript, closing bracket,
-      # or dot should have a space after it.
+      # dot, or a binary op should have a space after it.
       return True
```
harahu: Loosing coverage on this line now. Suggestions for a test?

gwelymernans: It's weird, because the coverage should have gone up not down. It's "only" a 0.03% loss, so we can ignore it. But if the line is no longer being executed that's more troubling. Could you do a quick check to make sure that it is being executed when you run `python setup.py test`?

harahu: The reason I loose coverage is because I made the conditional stricter. There are some `rval`s that were `not in '[)]}.'`, but they were all binary ops, which mean they don't hit the line now, even though they used to. So what we need are tokens that could conceivably come after a string but are not filtered by the new conditional. Suggestions?

gwelymernans: Ah! I read the patch wrong. This code looks like it's executed in "Chromium" style. There is a synthetic test like this:

```
b = '0' ('hello')
```

I say "synthetic" because I don't know why anyone would want to do something like this...but it looks like it's legal Python, so...

The test cases in `reformatter_basic_test.py` should be set up to use Chromium style by default.

PR #679, opened 2019-02-21
Fix bugs of handling arrays in pre-commit.sh

yous:
PYTHON_FILES, CHANGED_FILES are assigned by =$(...), so they're not arrays. This leads running `yapf -i -r "${PYTHON_FILES[@]}"` to be equivalent to running `yapf -i -r "$PYTHON_FILES"`, and that keeps yapf from handling multiple files properly.

Note that `${#ARRAY[@]}` gives the number of elements in `ARRAY`.

Regarding the way of initializing an array, `git diff --name-only` gives one filename per line. So reading line by line and append to the array is more solid even if the filename contains spaces.

You can check that if there is `foo bar.py`, then the previous implementation gives `PYTHON_FILES` with `foo` and `bar.py`.

Ref: https://github.com/koalaman/shellcheck/wiki/SC2207

### DIFF
--- a/plugins/pre-commit.sh
+++ b/plugins/pre-commit.sh
@@ -21,8 +21,10 @@
 # is used.
 
 # Find all staged Python files, and exit early if there aren't any.
-PYTHON_FILES=$(git diff --name-only --cached --diff-filter=AM | grep --color=never '.py$')
-if [ -z "${PYTHON_FILES}" ]; then
+PYTHON_FILES=()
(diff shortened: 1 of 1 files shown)


yous:
I've noticed that `read -d '' -r -a array < <(command)` always returns failure. Building an array from lines should be done by the loops.

See: https://github.com/koalaman/shellcheck/issues/1496

PR #690, opened 2019-03-14
Fix removing of extra lines on the edge of enabled lines boundaries.

kevincox:
This was broken by https://github.com/google/yapf/commit/dd9f72b9ec74267b84d5b0c0f70a12595f920166#diff-1020ef5a942231c7da662fa974a70a3bR145 which added a hacky fix that prevented removing newlines.

I haven't fixed the test that was added in that commit because I haven't yet work out how to tell if a range is affected by a disable comment. The correct solution will be to handle that comment instead of disallowing removal of lines.

Help welcome.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@
 - Parse integer lists correctly, removing quotes if the list is within a
   string.
 - Adjust the penalties of bitwise operands for '&' and '^', similar to '|'.
+- Re-enable removal of extra lines on the boundaries of formatted regions.
 
 ## [0.26.0] 2019-02-08
 ### Added
--- a/yapf/yapflib/reformatter.py
+++ b/yapf/yapflib/reformatter.py
@@ -114,12 +114,21 @@ def _RetainHorizontalSpacing(uwline):
 
 
 def _RetainRequiredVerticalSpacing(cur_uwline, prev_uwline, lines):
+  if cur_uwline.disable and (not prev_uwline or prev_uwline.disable):
+    # If both lines are disabled we aren't allowed to reformat anything.
+    lines = set()
+
   prev_tok = None
   if prev_uwline is not None:
     prev_tok = prev_uwline.last
   for cur_tok in cur_uwline.tokens:
     _RetainRequiredVerticalSpacingBetweenTokens(cur_tok, prev_tok, lines)
+
     prev_tok = cur_tok
+    if cur_uwline.disable:
+      # After the first token we are acting on a single line. So if it is
+      # disabled we must not reformat.
+      lines = set()
 
 
 def _RetainRequiredVerticalSpacingBetweenTokens(cur_tok, prev_tok, lines):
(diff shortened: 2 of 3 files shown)


kevincox:
cc @gwelymernans You introduced this issue.

gwelymernans:
Could you also update the `CHANGELOG`.

kevincox:
> Could you also update the CHANGELOG.

Done.

gwelymernans:
Great! Once you're ready, go ahead and send the PR.

kevincox:
I'm having trouble figuring out when lines are disabled via a comment in the code. It doesn't seem like they are added or removed from lines so there must be some other mechanism in play.

gwelymernans:
After the program is parsed and the unwrapped lines are created, it goes through to determine if there are any disable comments. If so it marks the affected lines as disabled.

kevincox:
Ok, I have fixed the disable check. It is a little hacky but better then before.
What is does is if the current line and the previous like are disabled it removes the lines from the set of editable lines.

A better approach would be to remove disabled lines from the set of editable lines for the whole reformat but that would require more refactoring and would best be done in a different fix.

gwelymernans:
> Ok, I have fixed the disable check. It is a little hacky but better then before.
> What is does is if the current line and the previous like are disabled it removes the lines from the set of editable lines.
> 
> A better approach would be to remove disabled lines from the set of editable lines for the whole reformat but that would require more refactoring and would best be done in a different fix.

+1

Could you add some tests to `yapf_test.py`? That way you can use the `--lines=<lines>` option. Also a few to `reformatter_basic_test.py` for coverage. I want to make sure this doesn't regress. (I should have mentioned this before. Sorry about that.)

kevincox:
On Thu., Mar. 21, 2019, 23:20 Bill Wendling, <notifications@github.com>
wrote:

> Ok, I have fixed the disable check. It is a little hacky but better then
> before.
> What is does is if the current line and the previous like are disabled it
> removes the lines from the set of editable lines.
>
> A better approach would be to remove disabled lines from the set of
> editable lines for the whole reformat but that would require more
> refactoring and would best be done in a different fix.
>
> +1
>
> Could you add some tests to yapf_test.py? That way you can use the
> --lines=<lines> option. Also a few to reformatter_basic_test.py for
> coverage. I want to make sure this doesn't regress. (I should have
> mentioned this before. Sorry about that.)
>

What specifically would you like tested? I have added some tests and there
are tests for the disable case. Do you just want more variety of the same
sort of tests?

>

gwelymernans:
Just some of the examples that demonstrated this bug. The blank lines calculator tests are good, but it doesn't necessarily indicate that the reformatter won't still mess things up.

kevincox:
Between the existing cases in blank-lines and the test added in https://github.com/google/yapf/commit/dd9f72b9ec74267b84d5b0c0f70a12595f920166 the obvious cases are covered. I will see if I can think up a couple more cases when I get the chance.

gwelymernans:
Okay. We can merge this then. If you can't think of others then no worries...

Issue #691, opened 2019-03-19
Prefer not splitting up list


apbard:
```
[style]
based_on_style=chromium
split_before_named_assigns=False
```

When there are lists as arguments yapf seems to adopt a weird splitting: 
```python
def fooooooooooooo(aka=['aaaaaaaa', 'aaaaaaa', 'aaaa', 'aa',
                        'a'], types=['double', 'bool'],
                   validator=some_looooooooooong_name):
    pass


def fooooooooooooo(type01=aaaaaa, type02=['double'], type03=['double'],
                   aka=['aaaaaaaa', 'aaaaaaa', 'aaaa', 'aa',
                        'a'], aaaaa=['double', 'bool'],
                   validator=some_looooooooooong_name):
    pass
```

instead I would prefer:
```python
def fooooooooooooo(aka=['aaaaaaaa', 'aaaaaaa', 'aaaa', 'aa', 'a'], 
                   types=['double', 'bool'],
                   validator=some_looooooooooong_name):
    pass


def fooooooooooooo(type01=aaaaaa, type02=['double'], type03=['double'],
                   aka=['aaaaaaaa', 'aaaaaaa', 'aaaa', 'aa', 'a'], 
                   aaaaa=['double', 'bool'],
                   validator=some_looooooooooong_name):
    pass
```

PR #692, opened 2019-03-19
ENH: slightly prefer split before named assigns

apbard:
Following up #691, this is an attempt to fix it by slightly favour split before named assign.
In the current state it only mitigates the effect in some cases:

```python
# now ok
def fooooooooooooo(aka=['aaaaaaaa', 'aaaaaaa', 'aaaa', 'aa', 'a'],
                   types=['double', 'bool'],
                   validator=some_looooooooooong_name):
    pass

# still bad
def fooooooooooooo(type01=['double'], type02=['double'], type03=['double'],
                   aka=['aaaaaaaa', 'aaaaaaa', 'aaaa', 'aa',
                        'a'], AAAAA=['double', 'bool'],
                   validator=some_looooooooooong_name):
    pass
``` 
Also it breaks a couple of tests that are right the way they are now. Example:
```python
  def f():
    if True:
      pytree_utils.InsertNodesBefore(
          _CreateCommentsFromPrefix(
              comment_prefix, comment_lineno, comment_column,
              standalone=True), ancestor_at_indent)  
```

@gwelymernans Any idea where/how to fix this?

### DIFF
--- a/yapf/yapflib/unwrapped_line.py
+++ b/yapf/yapflib/unwrapped_line.py
@@ -526,6 +526,7 @@ def _SplitPenalty(prev_token, cur_token):
   """Return the penalty for breaking the line before the current token."""
   pval = prev_token.value
   cval = cur_token.value
+
   if pval == 'not':
     return split_penalty.UNBREAKABLE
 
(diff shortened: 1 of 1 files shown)


gwelymernans:
Look around `format_decision_state.py:341`. That bit of code is trying to split before all named assigns. That bit of code probably isn't triggered for the cases above for some reason (could just be an option isn't set). I think you'll find that either you can modify that bit of code or gain some insight from it on how to address this.

apbard:
I am not sure we can do that there. Because ideally what we would like to achieve is something like:
"if we have to split somewhere, it is better to split before an argument than in the middle of a list"

PR #698, opened 2019-04-02
Update README.rst

jeongukjae:
* added links to similar/base projects in the section of Introduction.
* removed trailing whitespaces.
* corrected `in-place` to `in_place`. (in the context of explanation about python code)
* wrapped words related to code and file names with double backquotes.
* wrote detailed explanation about global config.

Q. is it right what I wrote in line 190? (I read codes related to that and tested but I feel it's uncertain)

> 4. In the `[style]` section of a `~/.config/yapf/style` file in your home
>   directory.


gwelymernans:
The comment on 190 seems correct to me. Thanks!

PR #710, opened 2019-04-16
Add knob SPLIT_ALL_TOP_LEVEL_COMMA_SEPARATED_VALUES

sergiogiro:
The existing knob SPLIT_ALL_COMMA_SEPARATED_VALUES is overly aggresive in that, if an argument list (or, more generally, a container) needs splitting, then all subexpressions will be split as well. For instance, in
```
abcdef(aReallyLongThing, b=(c,d))
```
if the line needs splitting because of `aReallyLongThing`, then it will produce
```
abcdef(
    aReallyLongThing,
    b=(c,
       d))
```
I've seen terrible things like
```
   argument: [Int,
              Int]
```
in function definitions with many arguments (even if the second `Int` fits in the line).

The new knob checks if a container (argument list, list literal, etc) fits in a line and avoids breaking if so, producing
```
abcdef(
    aReallyLongThing,
    b=(c, d))
```
which makes more sense (to me at least).


sergiogiro:
Sorry, meant to create the PR in my fork :S

PR #713, opened 2019-04-17
Add knob SPLIT_ALL_TOP_LEVEL_COMMA_SEPARATED_VALUES

sergiogiro:
Add knob SPLIT_ALL_TOP_LEVEL_COMMA_SEPARATED_VALUES

The existing knob SPLIT_ALL_COMMA_SEPARATED_VALUES is overly aggresive in that, if an argument list (or, more generally, a container) needs splitting, then all subexpressions will be split as well. For instance, in
```
abcdef(aReallyLongThing, b=(c,d))
```
if the line needs splitting because of `aReallyLongThing`, then it will produce
```
abcdef(
    aReallyLongThing,
    b=(c,
       d))
```
I've seen terrible things like
```
   argument: [Int,
              Int]
```
in function definitions with many arguments (even if the second `Int` fits in the line).

The new knob checks if a container (argument list, list literal, etc) fits in a line and avoids breaking if so, producing
```
abcdef(
    aReallyLongThing,
    b=(c, d))
```
which makes more sense (to me at least).

See here for a codebase moved from using split_all_comma_separated_values to split_all_top_level_comma_separated_values in a non-trivial codebase:
https://github.com/prodo-ai/plz/pull/249/files
This file is particularly illustrative.
https://github.com/prodo-ai/plz/pull/249/files#diff-fad3447da9b11c4372ab8fb7754675b2
You can see how it saves a lot of splits here:
https://github.com/prodo-ai/plz/pull/249/files#diff-1174ab209a5ec95fcc62ba032a05596eL124

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,11 @@
 # This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [0.27.1] UNRELEASED
+### Added
+- New knob `SPLIT_ALL_TOP_LEVEL_COMMA_SEPARATED_VALUES` is a variation on
+  `SPLIT_ALL_COMMA_SEPARATED_VALUES` in which, if a subexpression with comma
+  fits in its starting line, then the subexpression is not split (thus avoiding
+  unnecessary splits).
 ### Changed
 - Set `INDENT_DICTIONARY_VALUE` for Google style.
 ### Fixed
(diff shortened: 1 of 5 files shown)


gwelymernans:
This looks good. Could you update the `CHANGELOG` and `README.rst` files?

PR #738, opened 2019-07-10
Set JOIN_MULTIPLE_LINES = False in google style

garymm:
IIUC, this is more likely to make the output compliant with the google
style guide[1], though in some cases it is stricter than that
document requires.
Specifically this example from the "No" section would be fixed only
after this change:
```
if foo: bar(foo)
else: baz(foo)
```

But this example from the "Yes" section will also be modified:
```
if foo: bar(foo)
```

I think eliminating one way in which the formatter produes
style-guide-violating code is an improvement.

[1] https://github.com/google/styleguide/blob/gh-pages/pyguide.md#314-statements

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,7 @@
   unnecessary splits).
 ### Changed
 - Set `INDENT_DICTIONARY_VALUE` for Google style.
+- Set `JOIN_MULTIPLE_LINES = False` for Google style.
 ### Fixed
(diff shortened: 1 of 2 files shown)


gwelymernans:
I agree. Please add an entry in the `CHANGELOG`.

garymm:
> Please add an entry in the `CHANGELOG`.

Done.

PR #739, opened 2019-07-11
Patch for the error for the issue 730

priyansh19:


priyansh19:
Ok, Wait I will update the changes soon.

PR #776, opened 2019-10-31
stops the monkeypatch on lib2to3 grammar

moore3071:
Lib2to3 provides a proper grammar without exec statements for Python 3 come Python 3.7.5.
Since Python 3.7.5 was just released, it might be wise to delay this pull request for a brief amount of time so those using system-based package managers can catch up.

### DIFF
--- a/yapf/yapflib/pytree_utils.py
+++ b/yapf/yapflib/pytree_utils.py
@@ -81,13 +81,7 @@ def LastLeafNode(node):
   return LastLeafNode(node.children[-1])
 
 
-# lib2to3 thoughtfully provides pygram.python_grammar_no_print_statement for
-# parsing Python 3 code that wouldn't parse otherwise (when 'print' is used in a
-# context where a keyword is disallowed).
-# It forgets to do the same for 'exec' though. Luckily, Python is amenable to
-# monkey-patching.
-_GRAMMAR_FOR_PY3 = pygram.python_grammar_no_print_statement.copy()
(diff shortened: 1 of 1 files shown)


gwelymernans:
Unfortunately this doesn't work for Python 2, so it'll fail if someone tries to use Python 2...

hartwork:
@bwendling since Python 2 support is out the door by now, does this change the evaluation of this pull request for 2023?

bwendling:
@hartwork We could probably be more brutal with removing Python 2 from YAPF. For instance, we probably don't need the `PY3` variable anymore either. This also needs to go into the `CHANGELOG` and `README` files.

bwendling:
I'm not yet comfortably removing 3.7 support as there's likely to be 3.7 code still in use in a lot of places (e.g. Google).

As for updating the CHANGELOG, I know it's a hassle, but I feel that if we stop updating it with each commit then it'll be that much harder to gather the documentation for a release.

hartwork:
> I'm not yet comfortably removing 3.7 support as there's likely to be 3.7 code still in use in a lot of places (e.g. Google).

@bwendling alright, it's not dead yet. Now's a great time to start wearing a "Python 3.7 is almost dead!" shirt at work :smiley: 


> As for updating the CHANGELOG, I know it's a hassle, but I feel that if we stop updating it with each commit then it'll be that much harder to gather the documentation for a release.

Alright! :+1:

hartwork:
@bwendling can we close this as superseded by #1090?

Issue #622, opened 2018-09-25
[breaks user code] async for else: block


Zopieux:
yapf 0.24.0 (did not test earlier versions) introduces a syntax error in user code by wrongly indenting a statement if there is an `else:` block following an `async for` loop. This syntax is [legal](https://docs.python.org/3.5/reference/compound_stmts.html#the-async-for-statement) since 3.5 (where `async` and `await` were introduced). For reference, Intellij correctly handles this.

Input:
```python
async def test(  ):              
    async for item in gen(  ):
        print(  item  )
    else:               
        print(  "no result"  )
```

Expected output:
```python
async def test():
    async for item in gen():
        print(item)
    else:
        print("no result")
```

yapf 0.24.0 output:
```python
async def test():
    async for item in gen():
        print(item)
        else:  # syntax error
        print("no result")
```

Kaali:
Still broken in 0.28.0

Zopieux:
:man_facepalming: just use psf/black I guess.

gwelymernans:
FYI: https://github.com/google/yapf/pull/779

PR #779, opened 2019-11-13
Fix async for else indentation

Kaali:
Previously:

    async def fn():
        async for message in websocket:
            pass
        else:
            pass

was indented as:

    async def fn():
        async for message in websocket:
            pass
            else:
            pass

which breaks the code and crashes any further runs of yapf on that
codebase because of parse errors.

Fixes issue #622

### DIFF
--- a/yapf/yapflib/pytree_unwrapper.py
+++ b/yapf/yapflib/pytree_unwrapper.py
@@ -224,6 +224,8 @@ class PyTreeUnwrapper(pytree_visitor.PyTreeVisitor):
       if pytree_utils.NodeName(child) == 'ASYNC':
         break
     for child in node.children[index].children:
+      if (child.type == grammar_token.NAME and child.value == 'else'):
+        self._StartNewLine()
       self.Visit(child)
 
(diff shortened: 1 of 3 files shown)


Kaali:
@googlebot I signed it!

yapf/yapflib/pytree_unwrapper.py
```diff
@@ -224,6 +224,8 @@ def Visit_async_stmt(self, node):  # pylint: disable=invalid-name
       if pytree_utils.NodeName(child) == 'ASYNC':
         break
     for child in node.children[index].children:
+      if (child.type == grammar_token.NAME and child.value == 'else'):
```
gwelymernans: I believe you should be able to just use this (please check that `ELSE` is correct):

```
if pytree_utils.NodeName(child) == 'ELSE':
```

Kaali: The node name is `NAME`, and there doesn't seem to be a name for 'ELSE', so I would still have to check the value. Which as far as I understand it would make the line:

```
if (pytree_utils.NodeName(child) == 'NAME' and child.value == 'else'):
```

PR #789, opened 2019-12-30
Update chromium style to match what's used in Chromium

agrieve:
Adds a "YAPF" style in order to avoid reformatting the YAPF codebase
and updating test expectations

Relevant Chromium style discussion:
https://groups.google.com/a/chromium.org/g/chromium-dev/c/RcJgJdkNIdg/discussion

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -178,8 +178,8 @@ with a ``[yapf]`` heading. For example:
 
 The ``based_on_style`` setting determines which of the predefined styles this
 custom style is based on (think of it like subclassing). Four
-styles are predefined: ``pep8`` (default), ``chromium``, ``google`` and 
-``facebook`` (see ``_STYLE_NAME_TO_FACTORY`` in style.py_).
+styles are predefined: ``pep8`` (default), ``google`` and ``facebook`` (see
+``_STYLE_NAME_TO_FACTORY`` in style.py_).
 
 .. _style.py: https://github.com/google/yapf/blob/master/yapf/yapflib/style.py#L445
 
@@ -188,9 +188,9 @@ example:
 
 .. code-block:: shell
 
-    --style='{based_on_style: chromium, indent_width: 4}'
+    --style='{based_on_style: pep8, indent_width: 2}'
 
-This will take the ``chromium`` base style and modify it to have four space
+This will take the ``pep8`` base style and modify it to have two space
 indentations.
 
 YAPF will search for the formatting style in the following manner:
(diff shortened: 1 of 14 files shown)


agrieve:
Following chromium code review comments (switched fully to PEP8 since 79 vs 80 characters was the only difference), I've change the CL to just remove "chromium" as a separate style.

Added CHANGELOG entry.

README.rst
```diff
@@ -178,8 +178,8 @@ with a ``[yapf]`` heading. For example:
 
 The ``based_on_style`` setting determines which of the predefined styles this
 custom style is based on (think of it like subclassing). Four
-styles are predefined: ``pep8`` (default), ``chromium``, ``google`` and 
-``facebook`` (see ``_STYLE_NAME_TO_FACTORY`` in style.py_).
+styles are predefined: ``pep8`` (default), ``google`` and ``facebook`` (see
```
gwelymernans: Please add ``yapf`` to the list of predefined styles.

atetubou:
I think it is not good to replace style name atomically.
This makes it bit hard to migrate existing .style.yapf's for chromium.
https://cs.chromium.org/search/?q=c:chromium+f:style.yapf&sq=package:chromium&type=cs

agrieve:
I'm curious what you think would work better? A warning? If we silently support "chromium" as a style, no one will migrate.

atetubou:
I think it is better to have at least one release which has both yapf and chromium style in
https://github.com/google/yapf/releases
So that we can migrate our style files while using such release.

agrieve:
I think style=yapf is a bit of an obscure thing to transition to. I was hoping projects would transition to style=pep8, which has been available for a long time.

If you want to not have the style change at all, I think it's probably best to transition to what style=yapf sets explicitly, which I believe is:

    based_on_style = google
    ALLOW_MULTILINE_DICTIONARY_KEYS = True
    ALLOW_SPLIT_BEFORE_DEFAULT_OR_NAMED_ASSIGN = False
    INDENT_DICTIONARY_VALUE = True
    INDENT_WIDTH = 2
    SPLIT_BEFORE_BITWISE_OPERATOR = True
    SPLIT_BEFORE_DOT = True
    SPLIT_BEFORE_EXPRESSION_AFTER_OPENING_PAREN = True

gwelymernans:
@atetubou is correct though. I shouldn't have removed a style without a deprecation and migration plan and adjustment period. I'm sorry about that. It's not a good user experience...

PR #809, opened 2020-02-02
Extracts print_help into its own function

brianmego:
While working on another patch, extracted this function out of habit. No reason not to commit it.

### DIFF
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -215,6 +215,7 @@ def main(argv):
       verbose=args.verbose)
   return 1 if changed and (args.diff or args.quiet) else 0
 
+
 def print_help(args):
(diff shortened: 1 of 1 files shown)


brianmego:
I think something is up with the coverage check.  I only changed the one line and it's definitely still tested.

yapf/__init__.py
```diff
@@ -226,6 +215,20 @@ def main(argv):
       verbose=args.verbose)
   return 1 if changed and (args.diff or args.quiet) else 0
 
+def print_help(args):
```
gwelymernans: It looks like the contents of this function are over-indented. You should use two spaces.

brianmego: RIght you are. Updated.

PR #820, opened 2020-03-04
Knob for spaces around subscript colon

wwade:
**SPACES_AROUND_SUBSCRIPT_COLON**

Use spaces around the subscript / slice operator.  For example:

```python
  my_list[1 : 10 : 2]
```

This Bool option (False by default) inserts spaces around colons when
they're used as a subscript / slice operator.

There's a potential interaction with `SPACE_INSIDE_BRACKETS`, so there's
an extra test case in TestsForSpacesAroundSubscriptColon to verify the
behaviour with both enabled.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,8 @@
 - Don't split between two-word comparison operators: "is not", "not in", etc.
 - New knob `SPACE_INSIDE_BRACKETS` to add spaces inside brackets, braces, and
   parentheses.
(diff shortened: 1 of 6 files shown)


wwade:
@gwelymernans mind having a look at this one?  It's the last piece for our internal style guide so that we can switch over to yapf :)

gwelymernans:
Thanks! Sorry for the delay.

wwade:
> Thanks! Sorry for the delay.

No problem at all.

PR #831, opened 2020-04-23
CHANGELOG: New features to "Added"; MINOR version

wwade:
The FORCE_MULTILINE_DICT, SPACE_INSIDE_BRACKETS,
SPACES_AROUND_SUBSCRIPT_COLON should have been in an Added section
instead of Fixed.

Per Semantic Versioning, this also implies that the next release is now
a new MINOR, 0.30.0 (instead of 0.29.1).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,7 +2,14 @@
 # All notable changes to this project will be documented in this file.
 # This project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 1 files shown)


gwelymernans:
I think this is fixed with recent changes. Thank you, though!

wwade:
Indeed it is :)  What a race!

Issue #844, opened 2020-06-23
Add new line after logical operator


Conchylicultor:
Currently, I don't think there is a way to have each condition element to have their own line.
Additionally, the current formatting do not respect the `split_before_logical_operator = False`.
All condition get clustered like:
```py
if (
    very_long_variable_name is not None and verrry_long_variable_name.field > 0
    or very_long_variable_name.is_debugfs
):
  pass
```
I would like to format my code as:
```py
if (
    very_long_variable_name is not None and
    verrry_long_variable_name.field > 0 or
    very_long_variable_name.is_debugfs
):
  pass
```

My `.config/yapf/style`:
```ini
[style]
based_on_style = google
indent_width = 2
coalesce_brackets = True
dedent_closing_brackets = True
split_before_dot = True
split_before_logical_operator = False
```

keli:
True. Adding a knob for that would be nice.

MHannila:
I'd also like this. Currently this is the biggest problem with us using yapf on our codebase, because there are so many long if checks that'd become very hard to read.

jesse-sony:
This is an issue for us as well. Looking at the code, though it might be somewhat interesting to implement. It would need a new knob for sure. Then I guess whenever a Logical operator token is found in MustSplit, you need to check the total line length (which I think means iterating over `next_token` until you find None. Does not appear to be an existing method for this or a way to access the LogicalLine from the token to get `.last`). If the total line length is over the desired column count, then return `True` for the operator. 

Now that only works for splitting AFTER the operator, however. So I guess if the style has SPLIT_BEFORE_LOGICAL_OPERATOR set, then we have to always do a one token seek ahead and check for a Logical operator and perform the check at that time to force the split before we hit the logical operator.

This will have the impact, I think, of prioritizing logical operator wrapping over other types of wrapping, which I think is desirable but I'm not fully sure. Would need to see how it behaves in real code to get a good feel for that.

jesse-sony:
That also doesn't take into account a long chain of logical operators, some of which are nested inside parens. I guess the calculation for the wrap of a logical operator should take into account the total length of the clause, if the operator is nested inside parens or the total length of the line if it isn't nested

Conchylicultor:
Yapf is mostly unmaintained. I would suggest you to switch to Black or Pyink (google version that supports 2-space indentation, single quote,...)

jesse-sony:
[...]

PR #853, opened 2020-08-17
#844 new line added after logical operator

swapnalshahil:


swapnalshahil:
@googlebot I signed it!

gwelymernans:
Thanks for the submission, but the code isn't correct Python, and it's not according to our style guild.

Issue #635, opened 2018-10-28
Blank lines after import


vryazanov:
Hi there,
is it possible to add two blank lines after import? Currently it adds only one and I cannot find any way to change it.

arareko:
This feature will be available in version `0.31.0` of YAPF

Issue #347, opened 2016-12-13
Knob to set blank lines near imports, classes, functions and methods


advance512:
`BLANK_LINE_BEFORE_NESTED_CLASS_OR_DEF` allows adding a blank line before nested classes or methods.

Ideally, a knob would allow setting a specified amount of blank lines near:

* Top-level imports
* Top-level classes and functions
* Classes
* Methods
* Top-level comments

Alternatively, offer a knob can allow "maximum blank lines", thus allowing blank lines after imports (for example). Right now they get formatted away.

gwelymernans:
This is possible. I'll mark it as an "enhancement". Thanks!

jamesbeith:
`yapf` and `isort` currently handle blank lines near Top-level imports and comments differently. `isort` is happy with this:

```
from foo import bar


# Public


def baz():
    pass
```

but `yapf` wants to remove a line between the import and the comment.

```
from foo import bar

# Public


def baz():
    pass
```

Irrespective of `isort`'s behaviour, it would be nice to control the number of spaces here so there's 2 (ideally configurable) both above and below the comment. Should behave the same when there's a constant too.

```
from foo import bar

QUX = 'quux'


# Public


def baz():
    pass
```

advance512:
I have the same "infinite formatting loop" in PyCharm.
In the JavaScript world, you have eslint --fix and you have WebStorm following your .eslintrc file - so there's no such issue.

This is one of the reasons I don't actually use `yapf` but rather a series of `unify`, `autopep8`, `flake8` and `pylint`. I guess I'll add `isort` too. :)

gwelymernans:
Adding the knobs shouldn't be too hard. I'll bump up the priority here.

dlopuch:
I would also add a request knobs around number of blank lines inside functions (currently always exactly 1) and inside dictionaries (always exactly 0).

Kinda like eslint's [no-multiple-empty-lines](https://eslint.org/docs/rules/no-multiple-empty-lines)'s `max` option.

Inside a function I'd like to have the ability to use 2 lines to break up different sections, and inside long dictionaries, I'd like to be able to group logical sets of key/values.

arareko:
A knob `BLANK_LINES_BETWEEN_TOP_LEVEL_IMPORTS_AND_VARIABLES` will be available in version `0.31.0` of YAPF

diegovalenzuelaiturra:
[...]

Issue #340, opened 2016-11-22
Allow two empty lines between imports and assignments


gforcada:
At @plone we created our own styleguide and decided that between imports and code there should be 2 blank lines, to clearly mark the end of imports and beginning of code.

yapf though eats one of those two blank lines, would it be possible to add a knob for that?

gforcada:
I noticed that it already puts 2 blanks lines for imports and classes/functions but it only keeps one blank line when there is an assignment, i.e.:

```
import os


TOP_LEVEL = True
```

gets reformatted to:

```
import os

TOP_LEVEL = True
```

arareko:
This feature will be available in version `0.31.0` of YAPF

PR #854, opened 2020-08-19
Support custom number of blank lines between top-level imports and variable definitions

arareko:
This patch implements support for having a custom number of blank lines (default = 1) between top-level imports and variable definitions.

This feature has been previously requested in issues:
- #340
- #347
- #635 

It is specially useful to enable compatibility with tools like `isort`.

### DIFF
--- a/yapftests/reformatter_basic_test.py
+++ b/yapftests/reformatter_basic_test.py
@@ -280,7 +280,8 @@ class BasicReformatterTest(yapf_test_helper.YAPFTest):
     try:
       style.SetGlobalStyle(
           style.CreateStyleFromConfig(
-              '{based_on_style: yapf, blank_lines_after_top_level_imports: 2}'))
+              '{based_on_style: yapf, blank_lines_between_top_level_imports_and_variables: 2}'
+          ))
       uwlines = yapf_test_helper.ParseAndUnwrap(unformatted_code)
       self.assertCodeEqual(expected_formatted_code,
                            reformatter.Reformat(uwlines))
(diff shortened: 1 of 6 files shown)


yapftests/reformatter_basic_test.py
```diff
@@ -253,6 +253,43 @@ def f(  # Intermediate comment
     uwlines = yapf_test_helper.ParseAndUnwrap(unformatted_code)
     self.assertCodeEqual(expected_formatted_code, reformatter.Reformat(uwlines))
 
+  def testBlankLinesAfterTopLevelImport(self):
```
gwelymernans, line 256: Please add tests where you set the new knob to a non-default value.

arareko:
> How will this work if the line after an import isn't a variable definition? The wording of this knob makes it seem like it will change the blank lines after imports even if the next line is a class or function definition, which is wrong. The wording of the option should be changed to clarify this.

@gwelymernans how about `BLANK_LINES_BETWEEN_TOP_LEVEL_IMPORTS_AND_VARIABLES`?

arareko:
Updated with new wording for knob. Also added a couple more tests to verify the previous behaviour for classes and function definitions is respected.

gwelymernans:
Thanks!

yapftests/reformatter_basic_test.py
```diff
@@ -421,8 +510,8 @@ def testSingleComment(self):
 
   def testCommentsWithTrailingSpaces(self):
     unformatted_code = textwrap.dedent("""\
-        # Thing 1
```
JessicaNgo-CG, line 424: I think this change was unintended here (and might defeat the purpose of this test)

arareko: yes. seems like my IDE lacks the context of whats happening here.

arareko: fixed in #857

PR #860, opened 2020-09-10
Simplify _GRAMMAR_FOR_PY3

Bocete:
Use python_grammar_no_print_and_exec_statement directly instead of removing `exec` from python_grammar_no_print_statement. See https://github.com/python/cpython/blob/master/Lib/lib2to3/pygram.py#L39-L40

### DIFF
--- a/yapf/yapflib/pytree_utils.py
+++ b/yapf/yapflib/pytree_utils.py
@@ -81,13 +81,7 @@ def LastLeafNode(node):
   return LastLeafNode(node.children[-1])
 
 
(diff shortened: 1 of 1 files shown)


gwelymernans:
Please address the CI errors.

Bocete:
Thanks Bill. I'm more than bit out of my depth here, but it looks like not all of these Python binaries used by CI are before the fix for https://bugs.python.org/issue23896, apart from the nightly. Although I find that really surprising especially for 3.7, as that issue was closed over a year ago.

I'm inclined to close this issue as it's just minor cleanup and not really worth pursuing.

PR #864, opened 2020-09-18
Allow `lines` parameter to fix whitespace between statements.

msuozzo:
The previous logic largely ignored `lines` values that were not part of
statements. This means that yapf required a statement to be formatted to
make whitespace-only changes.

This logic tweak allows `lines` to be taken into account even when the
adjacent statements are disabled (i.e. not in `lines`).


msuozzo:
Coverage doesn't really look like it decreased. Could have just been the result of removing some code.

PR #870, opened 2020-10-16
Add yapf-diff command

### LINKED ISSUE #190: yapf-diff (similar to clang-format-diff) ?
I am using clang-format a lot. And it's very convenient to do so via clang-format-diff script (quick description from https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/clang-format-diff.py):

```
ClangFormat Diff Reformatter
============================

This script reads input from a unified diff and reformats all the changed
lines. This is useful to reformat all the lines touched by a specific patch.
Example usage for git/svn users:

  git diff -U0 HEAD^ | clang-format-diff.py -p1 -i
  svn diff --diff-cmd=diff -x-U0 | clang-format-diff.py -i
```

It would great to have similar script for yapf. Looks like currently yapf doesn't have this mode available. 

Do you have plans on supporting this?

bouweandela:
This could be a way to add a `yapf-diff` command. I took the latest copy of [clang-format-diff.py](https://github.com/llvm/llvm-project/blob/master/clang/tools/clang-format/clang-format-diff.py), took the required changes from this gist https://github.com/google/yapf/issues/190#issuecomment-285295786, tried to align the command line arguments with the `yapf` arguments and formatted the result with yapf 0.30.0.

Does this look reasonable, or would you like a more integrated argument in the existing `yapf` command?

I would be happy to add some unit tests if this looks like something that could be merged.

Closes #190

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -321,6 +321,42 @@ The ``in_place`` argument saves the reformatted code back to the file:
     >>> print(open("foo.py").read())  # contents of file (now fixed)
     a == b
 
+Formatting diffs
+================
+
+Options::
+
+    usage: yapf-diff [-h] [-i] [-p NUM] [--regex PATTERN] [--iregex PATTERN][-v]
+                     [--style STYLE] [--binary BINARY]
+
+    This script reads input from a unified diff and reformats all the changed
+    lines. This is useful to reformat all the lines touched by a specific patch.
+    Example usage for git/svn users:
+
+      git diff -U0 --no-color --relative HEAD^ | yapf-diff -i
+      svn diff --diff-cmd=diff -x-U0 | yapf-diff -p0 -i
+
+    It should be noted that the filename contained in the diff is used
+    unmodified to determine the source file to update. Users calling this script
+    directly should be careful to ensure that the path in the diff is correct
+    relative to the current working directory.
+
+    optional arguments:
+      -h, --help            show this help message and exit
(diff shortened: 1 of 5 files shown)


gwelymernans:
This looks good. It needs a bit of cleanup. The top-level comment in yapf-diff shouldn't be about clang-diff, and it needs the copyright header. Should also be mentioned in CHANGES.

gwelymernans:
Hold on. Ignore what I said about the header stuff. Copyright stuff is more complex. :-) I'll need to check about whether we can do this.

bouweandela:
Thanks! Do you have any news about the legal question already? The alternative would be to implement something from scratch, but because the script is fairly simple the difference between the two options might be pretty small in the end.

gwelymernans:
> Thanks! Do you have any news about the legal question already?

I've gotten the info on how to proceed. We need to ensure that the licensing for this tool is retained. So please add it to a subdirectory called `third_party`. The structure will look something like this:

```
yapf/
  |
  ` third_party/
    |
    ` yapf_diff/
      |
      - LICENSE
      |
      - ... (whatever files belong here)
```

The `LICENSE` file should have a copy of the full license text under which clang-format was released. (Assuming this is clang's general license.) Because we're modifying the name, please add a comment at the top of `yapf-diff` that it's a modified copy of clang-format.

PR #881, opened 2020-12-02
Fix directory exclusion on Windows.

### LINKED ISSUE #880: Exluding directories fails on Windows due to \ vs / path separators.
Currently using `--exclude` on a folder, as well putting one in the `.gitignore`, has no effect. I've dug through and found the culprit:

`IsIgnored()` (below) does the filtering. It knows it must strip leading `./`s from the paths it tests before applying `fnmatch()`.

https://github.com/google/yapf/blob/768267f3481259df9e2162968c9d5986074ec15c/yapf/yapflib/file_resources.py#L187-L192

But on Windows, `os.walk()` (the output of which gets fed to `IsIgnored()`) returns Windows style `.\file\name` paths which fail to have the `.\` prefix stripped so `IsIgnored()` always return false.

The easiest fix IMO is just to normalise paths to Posix style paths, then leave the `./` stripping code as is:
```diff
diff --git i/yapf/yapflib/file_resources.py w/yapf/yapflib/file_resources.py
index 50b4687..477e4c1 100644
--- i/yapf/yapflib/file_resources.py
+++ w/yapf/yapflib/file_resources.py
@@ -186,6 +186,9 @@ def _FindPythonFiles(filenames, recursive, exclude):
 
 def IsIgnored(path, exclude):
   """Return True if filename matches any patterns in exclude."""
+  if os.name == "nt":
+    # Normalise "\" separators to "/" on Windows.
+    path = path.replace("\\", "/")
   path = path.lstrip('/')
   while path.startswith('./'):
     path = path[2:]
```

Shall I put the above into a pull request?

bwoodsend:
Following on from #880, swap some hard-coded `/`s for `os.path.sep` so that `--exclude folder` works on Windows. 

The 3rd commit is the main change. The 1st is just to get the corresponding `TestCase` to run on Windows. I'm slightly confused as to why the 2nd is needed on Windows but not Linux.

Closes #880.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,8 @@
 - Add 'BLANK_LINES_BETWEEN_TOP_LEVEL_IMPORTS_AND_VARIABLES' to support setting
   a custom number of blank lines between top-level imports and variable
   definitions.
+### Fixed
+- Exclude directories on Windows.
 
 ## [0.30.0] 2020-04-23
 ### Added
--- a/yapf/yapflib/file_resources.py
+++ b/yapf/yapflib/file_resources.py
@@ -154,6 +154,7 @@ def _FindPythonFiles(filenames, recursive, exclude):
   """Find all Python files."""
   if exclude and any(e.startswith('./') for e in exclude):
     raise errors.YapfError("path in '--exclude' should not start with ./")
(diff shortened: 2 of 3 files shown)


bwoodsend:
@googlebot I signed it!

bwoodsend:
That cla bot is cool.

PR #886, opened 2020-12-18
Ignore end of line `# copybara:` directives.

gpshead:
pylint and pytype end of line comments adjusting their behavior are
already handled specially.  This does the same for copybara.

I don't like that this is turning into a chain of property calls,
refactoring this into a single question of "is this end of line comment
special" would be better if we gain any more of these.

A usual internal comment this will ignore is `# copybara:strip` placed
on a long import line for something being excluded by copybara when
pushing internally maintained code to upstream to keep internal
google-isms from escaping.


gwelymernans:
Thanks. Could you add an entry to `CHANGELOG`?

gpshead:
> Could you add an entry to CHANGELOG?

Yep, done.

Issue #708, opened 2019-04-15
Integration with pyproject.toml for Formatting style


samypr100:
[PEP 518](https://www.python.org/dev/peps/pep-0518/) defines `pyproject.toml` as a configuration file for software packages to use to specify their build dependencies.

Particularly, it provides an alternative to `setup.cfg`.

For example, a section in `pyproject.toml` could look like:
```toml
[tool.yapf]
based_on_style = "pep8"
spaces_before_comment = 4
split_before_logical_operator = true
```

Initially, `yapf` could look for a `pyproject.toml` on the project and determine formatting style settings from that similar to how it does with `setup.cfg`.

This request would be an additional method to the currently supported methods of specifying `yapf` settings. Any chance we could see this in `yapf`?

JohnMoutafis:
Is this going to be implemented? 
If yes, what is the process state atm?

dclong:
Looking forward to this feature.

gwelymernans:
I don't see an issue with doing this. I don't have any knowledge about `pyproject.toml`, but I'll look into it. If someone wanted to submit a PR, we would accept it :)

dclong:
Does yapf support configure `.yapfignore` in pyproject.toml too?

hirosassa:
@dclong No, I implemented only [formatting style](https://github.com/google/yapf/blob/59a3b76db1319287be3d580512d7caf9ca4fbd4a/README.rst#id7) configuration in the PR above.

dclong:
@hirosassa Thank you for the info! I've opened an issue https://github.com/google/yapf/issues/904 requesting support yapf ignore configuration in pyproject.toml.

KuSi833:
I believe this issue can be closed since it was solved with PR #893?

hirosassa:
@KuSi833 Thanks! 

@gwelymernans Would you close this issue?

juharris:
Is there a way to specify knobs in pyproject.toml?

PR #893, opened 2021-01-27
pyproject.toml support

hirosassa:
This PR is for #708

I added `pyproject.toml` support.
Section syntax of `pyproject.toml` file is the same as @samypr100 's explanation in #708.

Please let me know if I miss anything.
Thank you.

### DIFF
--- a/yapf/yapflib/style.py
+++ b/yapf/yapflib/style.py
@@ -17,6 +17,8 @@ import os
 import re
 import textwrap
 
+import toml
+
 from yapf.yapflib import errors
 from yapf.yapflib import py3compat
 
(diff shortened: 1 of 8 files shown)


yapf/yapflib/style.py
```diff
@@ -17,6 +17,8 @@
 import re
 import textwrap
 
+import toml
```
hirosassa, line 20: [help wanted] I added external dependency because there's no toml loader in standard libraries. Is this OK?

hirosassa:
@gwelymernans Hi! This PR is ready for review. 
If miss anything, please let me know. 
Thank you!

yapf/yapflib/style.py
gwelymernans: I'd prefer this to be seamless. So if someone doesn't have the `toml` package installed, then we just skip this support all together. Is it possible to import the `toml` conditionally and if it fails to import set a flag or something?

gwelymernans (commented):
Sorry for the late review.

yapf/yapflib/style.py
hirosassa: I agree with you! Fixed this in 58b9034

PR #898, opened 2021-02-05
Prevent os.walk() scanning excluded dirs.

bwoodsend:
Rather than discovering every file then eliminating those in excluded folders, this eliminates searching the excluded folders to begin with.

This is a big speedup if you have an exlcuded virtual environment sat in a repository.


gwelymernans:
Thank you!

PR #921, opened 2021-04-26
Fix yapf-diff install by automatically finding valid packages

### LINKED ISSUE #920: Yapf-diff not installed correctly
To reproduce, on a clean system:

```
[me@localhost] $ pip install -U yapf
[me@localhost] $ yapf-diff -h
  File "/Users/me/.pyenv/versions/3.7.10/bin/yapf-diff", line 5, in <module>
    from yapf.third_party.yapf_diff.yapf_diff import main
ModuleNotFoundError: No module named 'yapf.third_party'
```

This is happening because of an error in the `setup.py` file: `yapf.third_party.yapf_diff` was not added to the `packages` kwarg to `setup()`, so the files are not actually getting installed.

I'm pushing a PR that fixes it in a minute.

Surprising no one has brought it up yet, but ¯\_(ツ)_/¯ . I'm just happy to see it fixed.

myw:
This PR fixes #920 by having `setup.py` automatically find all python packages within the tree. This avoids a similar issue in the future if other packages are added inside `third_party`. 

Only directories with `__init__.py` files will be treated as packages, so `plugins`, for example, is not searched. For the more conservative, this could be refactored to only recursively search `third_party`, and require explicit lists everywhere else.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ import codecs
 import sys
 import unittest
 
-from setuptools import setup, Command
+from setuptools import find_packages, setup, Command
 
 import yapf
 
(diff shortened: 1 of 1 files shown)


myw:
@googlebot I signed it!

gwelymernans:
Thanks for the patch!

PR #938, opened 2021-07-12
Merge isinstance calls

monosans:


monosans:
@googlebot I signed it!

PR #952, opened 2021-08-13
Added yapf_api.FormatTree

### LINKED ISSUE #951: Allow formatting a parsed lib2to3 tree
We have a tool which applies transformations from lib2to3 before calling `yapf_api.FormatCode`. This leads to redundant tokenization and parsing on the boundary between lib2to3 and YAPF, even though both use the same CST internally. 

Would you accept a PR adding `yapf_api.FormatTree` which would do exactly what `yapf_api.FormatCode` does, but without the parsing step?

superbobry:
The new entry point allows to avoid re-parsing when a lib2to3 pytree has already been constructed.

Closes #951.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,8 @@
 ### Added
 - Look at the 'pyproject.toml' file to see if it contains ignore file information
   for YAPF.
+- New entry point `yapf_api.FormatTree`for formatting lib2to3 concrete
+  syntax trees.
 ### Fixed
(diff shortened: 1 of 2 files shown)


superbobry:
@gwelymernans is there anything else I need to do before to make this mergeable?

gwelymernans:
Sorry about that. I was away for a few days. Merged. Thank you!

Alnsim69 (commented):
yapf/yapflib/yapf_api.py

Linksforyou (approved):

PR #960, opened 2021-09-12
Update README.rst for style file format

NelsonMinar:
I'm a brand new user, so this could be wrong. The README says your style file puts style declarations in a `[yapf]` block. That doesn't work for me with yapf 0.31.0 from pip; it looks like it should be a `[style]` block. 

This change updates the README for new users like me copy-pasting. I noted there's still a few occurrences of `[yapf]` in the code. Mostly comments, but also one test. Maybe those should be updated too.


NelsonMinar:
I'm not signing a legal agreement to suggest a two line change to a documentation file. I'll just file a bug report.

PR #962, opened 2021-09-16
``pytree_utils.ParseCodeToTree`` now adds a trailing EOL if needed

superbobry:
This was previously done by ``yapf_api.FormatCode``, but I think it makes
more sense to have it in ``pytree_utils.ParseCodeToTree``, because the
grammar requires a trailing EOL.


gwelymernans:
LGTM. Thanks!

Issue #936, opened 2021-07-09
`toml` is missing as a dependency


stefsmeets:
Hi all,

I noticed that `toml` is not specified as a dependency. This causes issues when no configuration is specified in any of the search locations (i.e. in `setup.cfg`), while `pyproject.toml` is present: 
https://github.com/google/yapf/blob/c6077954245bc3add82dafd853a1c7305a6ebd20/yapf/yapflib/style.py#L747

The error message:
`yapf: toml package is needed for using pyproject.toml as a configuration file`

This causes an issue with the `pre-commit` hook, which installs `yapf` in its own virtual environment:
https://github.com/pre-commit/mirrors-yapf/issues/15

While it is possible to work around the issue by adding a dummy config to `setup.cfg` or `.style.yapf`, with the increased prominence `pyproject.toml` becoming the location to declare the build system, would it make sense to add `toml` and an installation requirement?

PR #964, opened 2021-09-20
Added pyproject extra to install toml package as an optional dependency

arareko:
This patch fixes https://github.com/google/yapf/issues/936

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 - New entry point `yapf_api.FormatTree` for formatting lib2to3 concrete
   syntax trees.
 ### Fixed
+- Added toml package as a dependency.
 - Enable `BLANK_LINE_BEFORE_NESTED_CLASS_OR_DEF` knob for "pep8" style, so
   method definitions inside a class are surrounded by a single blank line as
   prescribed by PEP8.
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
👎 Should be marked as an optional dependency.

arareko:
> 👎 Should be marked as an optional dependency.

@GrahamCampbell not optional when it is being used directly in one of the YAPF modules.

GrahamCampbell:
The code expects it to be missing and handles that case. Totally not the same as just straight up importing it.

gwelymernans:
I have to agree with @GrahamCampbell here.

dzshn:
I think this change is reasonable considering `pyproject.toml` is becoming standard for declaring build systems, but also what mentioned in #936.. or at least I do find it unusual to have to add `toml` as a dep myself when using poetry with yapf

arareko:
@gwelymernans I have made the necessary adjustments.

pedro757:
Question: why toml can not just be a normal dependency?

I mean, in the end **yapf** is just a **developer dependency** it actually wont affect any program performance.

pfernandesviz:
Can you guys merge this?

gwelymernans:
Sorry about the lateness. Done.

PR #970, opened 2021-10-07
fix: fix tab-space mix with tab indenation and continuation_align_style=SPACE

### LINKED ISSUE #967: use_tabs=True causes tabs-space mix
```
[style]
based_on_style=pep8
use_tabs=True
column_limit = 200 
```

with
```
class CLI(object):
»   def __init__(self):
»   »   self.modes = { 
»   »   »   'request': self.request,
»   »   »   'response': self.response,
»   »   »   'parse': { 
»   »   »   »   'request': self.parse_request,
»   »   »   »   'response': self.parse_response,
»   »   »   }
»   »   }
```

causes tab-space mix:
```
class CLI(object):
»   def __init__(self):
»   »   self.modes = {
»   »       'request': self.request,
»   »       'response': self.response,
»   »       'parse': {
»   »           'request': self.parse_request,
»   »           'response': self.parse_response,
»   »       }
»   »   }
```

note: `»` represents a tab.

spaceone:
Fixes #967 

Let's see what the tests say about this?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,8 @@
   method definitions inside a class are surrounded by a single blank line as
   prescribed by PEP8.
 - Fixed the '...' token to be spaced after a colon.
+- With `USE_TABS=True` and `continuation_align_style=SPACE` no space tab mix
+  is produced anymore.
 
 ## [0.31.0] 2021-03-14
 ### Added
--- a/yapf/yapflib/format_token.py
+++ b/yapf/yapflib/format_token.py
@@ -77,7 +77,9 @@ def _TabbedContinuationAlignPadding(spaces, align_style, tab_width):
     if spaces > 0:
       return '\t' * int((spaces + tab_width - 1) / tab_width)
     return ''
-  return ' ' * spaces
+  if tab_width > 1 and spaces % tab_width == 0:
+    return '\t' * (spaces // tab_width)
+  return ' ' * spaces  # TODO: add a comment why we indent with spaces in a tab area?
 
 
 class FormatToken(object):
--- a/yapftests/format_token_test.py
+++ b/yapftests/format_token_test.py
@@ -30,7 +30,7 @@ class TabbedContinuationAlignPaddingTest(unittest.TestCase):
     self.assertEqual(pad, '')
 
     pad = format_token._TabbedContinuationAlignPadding(2, align_style, 2)
-    self.assertEqual(pad, ' ' * 2)
+    self.assertEqual(pad, '\t')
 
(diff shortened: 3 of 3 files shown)


gwelymernans:
I'm sorry to get to this so late. Could you update the CHANGELOG please? Also, please add a test case derived from the original issue.

spaceone:
> I'm sorry to get to this so late. Could you update the CHANGELOG please? Also, please add a test case derived from the original issue.

OK, I created a changelog entry.
Is a new test case necessary? As I adjusted a current one, which already tests this.

gwelymernans:
I would prefer the testcase just so we have a concrete example of what we're fixing, so that if something goes wrong we won't mess it up.

dhouck:
(To clarify: Iʼm a member of the general public, not Google)

I think what you want is best solved with just using `CONTINUATION_ALIGN_STYLE=FIXED`; changing what `CONTINUATION_ALIGN_STYLE=SPACE` does will break things.

The point of using spaces for alignment is so that something like
```python
def foo_function(arg1, arg2,
                 arg3):
»   return [
»           'hello',
»           'world',
»   ]
```
looks good (ie. the `arg3` right under the `arg1` and the strings right after the `[`) even when viewed with a different tab width.  However, as you see in #967, if it canʼt figure out how to line up continuation lines, itʼll just fall back to four spaces, which isnʼt ideal.

Maybe a better simple option would be to have the default `CONTINUATION_ALIGN_STYLE` be `SPACE` when `USE_TABS` is false, and `FIXED` when itʼs true?  It looks like itʼd take a bigger refactor to add a fourth `CONTINUATION_ALIGN_STYLE` that uses tabs for generic continuations and spaces if it can actually align things, I think?

spaceone:
Since Python3 mixing tabs with spaces is a SyntaxError - so we `USE_TABS=True` ist must make sure that spaces are never used for indentation.

dhouck:
Putting spaces *before* a tab is an error.  Putting spaces after a tab is not, and even if it were this only uses spaces for continuation lines, not technically indentation, where it definitely isnʼt an error.  I just tested with my example code and with
```python
def foo2():
»   print("Unconditional")
»   if foo_function(1,2,3):
»       print("Conditional")
````
which uses tabs and then spaces for *indentation*, not just the alignment yapf generates, and both work in python 3.9.7.

And if it were an error, then this PR wouldnʼt help, because not all indentations are a multiple of `tab_width`; youʼd need one of the other two (already existing) options for `CONTINUATION_ALIGN_STYLE`.

Issue #983, opened 2021-12-16
match case statement raise lib2to3.pgen2.parse.ParseError: bad input


heavenshell:
Hi.

Using Python3.10 match case raise exception.
Version: yapf-0.31.0

```python
def http_error(status):
    match status:
        case _:
            return ""
```

```
Traceback (most recent call last):
  File "tmp/.venv/lib/python3.10/site-packages/yapf/yapflib/pytree_utils.py", line 115, in ParseCodeToTree
    tree = parser_driver.parse_string(code, debug=False)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/lib2to3/pgen2/driver.py", line 103, in parse_string
    return self.parse_tokens(tokens, debug)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/lib2to3/pgen2/driver.py", line 71, in parse_tokens
    if p.addtoken(type, value, (prefix, start)):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/lib2to3/pgen2/parse.py", line 162, in addtoken
    raise ParseError("bad input", type, value, context)
lib2to3.pgen2.parse.ParseError: bad input: type=1, value='status', context=(' ', (2, 10))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/tmp/.venv/bin/yapf", line 8, in <module>
    sys.exit(run_main())
  File "tmp/.venv/lib/python3.10/site-packages/yapf/__init__.py", line 362, in run_main
    sys.exit(main(sys.argv))
  File "tmp/.venv/lib/python3.10/site-packages/yapf/__init__.py", line 126, in main
    changed = FormatFiles(
  File "tmp/.venv/lib/python3.10/site-packages/yapf/__init__.py", line 204, in FormatFiles
    changed |= _FormatFile(filename, lines, style_config, no_local_style,
  File "tmp/.venv/lib/python3.10/site-packages/yapf/__init__.py", line 227, in _FormatFile
    reformatted_code, encoding, has_change = yapf_api.FormatFile(
  File "tmp/.venv/lib/python3.10/site-packages/yapf/yapflib/yapf_api.py", line 94, in FormatFile
    reformatted_source, changed = FormatCode(
  File "tmp/.venv/lib/python3.10/site-packages/yapf/yapflib/yapf_api.py", line 147, in FormatCode
[...]

PR #986, opened 2021-12-30
Add support for Python 3.10

hugovk:
And updates for lower Python 3.x supported version.


gwelymernans:
Thanks! Please add an update in CHANGELOG.

hugovk:
CHANGELOG updated!

wookayin:
The current (master) version of yapf still does not support 3.10 syntax, see #983. CHANGELOG and this PR sounds like yapf now fully supports python 3.10, but it's indeed not the case yet...

daskol:
Could you add a release tag e.g 0.33.0. I guess that support to Python 3.10 is weighty reason to make a release. Also, there are no updates for package in Linux distros because there is no release. 

P.S. I'd like to express my support to request #1001.

PR #1005, opened 2022-04-21
fix for indent block comments issue

### LINKED ISSUE #969: indent block comments but remove comment-indentation
```
[style]
based_on_style=pep8
use_tabs=True
```

transforms:
```
class Foo(object):
»   def __init__(self):
»   #»  pass
»   »   pass
```

into:
```
class Foo(object):
»   def __init__(self):
»   »   #»  pass
»   »   pass
```

I think (wrong indented) block comments should be indented, but the indentation in the comment should be removed.
So that we get: 

```
class Foo(object):
»   def __init__(self):
»   »   # pass
»   »   pass
```

jcipa:
Fixes #969

### DIFF
--- a/yapf/yapflib/reformatter.py
+++ b/yapf/yapflib/reformatter.py
@@ -62,6 +62,17 @@ def Reformat(llines, verify=False, lines=None):
 
     if not lline.disable:
       if lline.first.is_comment:
+        for character in lline.first.value:
+          if character == '#':
+            comment_before_tab = True
+          elif character == '\t':
+            break
+
+        if comment_before_tab:
+          non_comment_text_index = (len(lline.first.value[1:]) - len(lline.first.value[1:].lstrip())) + 1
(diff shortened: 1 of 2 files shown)


yapf/yapflib/reformatter.py
```diff
@@ -62,6 +62,17 @@ def Reformat(llines, verify=False, lines=None):
...
     if not lline.disable:
       if lline.first.is_comment:
+        for character in lline.first.value:
+          if character == '#':
+            comment_before_tab = True
+          elif character == '\t':
+            break
+
+        if comment_before_tab:
+          non_comment_text_index = (len(lline.first.value[1:]) - len(lline.first.value[1:].lstrip())) + 1
```
bwendling, line 72: 80-column formatting please.

bwendling (commented):
Please add a note to the `CHANGELOG`.

PR #1022, opened 2022-08-18
Add a new knob and a change for inline comment alignment

lizawang:
**Hello yapf founders!**
I want to share one knob and one change for the *AlignTrailingComments* feature.

### The knob: *ALIGN_NEWLINE_COMMENTS_WITH_INLINE_COMMENTS*
This is a style option I added for when we do not want to align the newline comments with the inline comments in the same alignment block.
- The default setting is **True**. When set to **False**, newline comments won't align with inline comments.
- For example, instead of 

  ```
  if 1:
  encoding_bom = None
  if ( encoding.endswith( '-sig' ) or encoding.endswith( '-bom' ) ):
    encoding = encoding[ :-4 ]
    if ( encoding == 'utf-8' ):          # comment inline
                                         # comment newline
      encoding_bom = codecs.BOM_UTF16_BE
    else:
                                         #=========Comment Newlines=============================
                                         # I18N_MESSAGE | 210620850
                                         # de | BOM nicht unterstützt für Encoding '${enc}'.
                                         # en | BOM not supported for encoding '${enc}'.
                                         #======================================================
      result[ 'success' ] = False
      result[ 'error_code' ] = 210620850 # comment inline
  ```
- we want
  ```
  if 1:
  encoding_bom = None
  if ( encoding.endswith( '-sig' ) or encoding.endswith( '-bom' ) ):
    encoding = encoding[ :-4 ]
    if ( encoding == 'utf-8' ):          # comment inline
      # comment newline
      encoding_bom = codecs.BOM_UTF16_BE
    else:
      #=========Comment Newlines=============================
      # I18N_MESSAGE | 210620850
      # de | BOM nicht unterstützt für Encoding '${enc}'.
      # en | BOM not supported for encoding '${enc}'.
      #======================================================
      result[ 'success' ] = False
      result[ 'error_code' ] = 210620850 # comment inline
  ```

### One change
Another change I added is when there is an object(dictionary, argument list, list, function call) with its entries on separate lines with its closing bracket also on a separate line,  it starts a new alignment block.
For example:
```
func( 1 )               # Line 1
func( 2 )               # Line 2
d = {
    key1: value1,
    key2: value2,
    key3: value3,
}                       # Line 3
func( 3 )     # Line 4
func( 4 )     # line 5
```
### Finally
- The knob is added in *style.py*. The default setting is **True**.
- The tests for the new knob and new alignment after the object with entries on separate lines are added in *yapf_test.py*.

<br>

### DIFF
--- a/yapf/pytree/comment_splicer.py
+++ b/yapf/pytree/comment_splicer.py
@@ -42,7 +42,6 @@ def SpliceComments(tree):
   # This is a list because Python 2.x doesn't have 'nonlocal' :)
   prev_leaf = [None]
   _AnnotateIndents(tree)
-
   def _VisitNodeRec(node):
     """Recursively visit each node to splice comments into the AST."""
     # This loop may insert into node.children, so we'll iterate over a copy.
--- a/yapf/yapflib/format_decision_state.py
+++ b/yapf/yapflib/format_decision_state.py
@@ -644,6 +644,12 @@ class FormatDecisionState(object):
 
     if not dry_run:
       indent_level = self.line.depth
+      #-----------------------NOTE-----------------------------------
+      # we want newline comments inside the lline to
+      # keep their original indentations
+      if current.is_comment:
+        indent_level = int(current.column/style.Get('INDENT_WIDTH'))
+      #---------------------Added by Xiao----------------------------
       spaces = self.column
       if spaces:
         spaces -= indent_level * style.Get('INDENT_WIDTH')
--- a/yapf/yapflib/format_decision_state.py
+++ b/yapf/yapflib/format_decision_state.py
@@ -644,12 +644,6 @@ class FormatDecisionState(object):
 
     if not dry_run:
       indent_level = self.line.depth
-      #-----------------------NOTE below-----------------------------------
-      # we want newline comments inside the lline to
-      # keep their original indentations
-      if current.is_comment:
-        indent_level = int(current.column/style.Get('INDENT_WIDTH'))
-      #---------------------Added by Xiao----------------------------
       spaces = self.column
       if spaces:
         spaces -= indent_level * style.Get('INDENT_WIDTH')
--- a/yapf/yapflib/format_decision_state.py
+++ b/yapf/yapflib/format_decision_state.py
@@ -644,7 +644,7 @@ class FormatDecisionState(object):
 
     if not dry_run:
       indent_level = self.line.depth
-      #-----------------------NOTE-----------------------------------
+      #-----------------------NOTE below-----------------------------------
       # we want newline comments inside the lline to
       # keep their original indentations
       if current.is_comment:
--- a/yapf/yapflib/reformatter.py
+++ b/yapf/yapflib/reformatter.py
@@ -277,7 +277,8 @@ def _AlignTrailingComments(final_lines):
     for tok in line.tokens:
       if (tok.is_comment and isinstance(tok.spaces_required_before, list) and
           tok.value.startswith('#')):
-        # All trailing comments and comments that appear on a line by themselves
+        # All trailing comments 
+        # NOTE not including comments that appear on a line by themselves
         # in this block should be indented at the same level. The block is
         # terminated by an empty line or EOF. Enumerate through each line in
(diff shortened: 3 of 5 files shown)


yapf/yapflib/format_decision_state.py
```diff
@@ -934,16 +934,19 @@ def _GetNewlineColumn(self):
...
 
+    cont_aligned_indent = self._IndentWithContinuationAlignStyle(
+        top_of_stack.indent)
+
     if isinstance(current.spaces_required_before, list):
-      # Don't set the value here, as we need to look at the lines near
-      # this one to determine the actual horizontal alignment value.
-      return 0
+      # only when the commet is not inside an object(list, dictionary,
+      # function call)logical line that is in many output lines
```
gwelymernans, line 942: Could you clarify this comment, please? I'm not sure what it's trying to suggest.

yapf/pytree/comment_splicer.py
```diff
@@ -42,7 +42,6 @@ def SpliceComments(tree):
   # This is a list because Python 2.x doesn't have 'nonlocal' :)
   prev_leaf = [None]
   _AnnotateIndents(tree)
-
```
gwelymernans, line 45: Please retain this newline.

yapf/yapflib/reformatter.py
```diff
@@ -277,7 +278,8 @@ def _AlignTrailingComments(final_lines):
     for tok in line.tokens:
       if (tok.is_comment and isinstance(tok.spaces_required_before, list) and
           tok.value.startswith('#')):
-        # All trailing comments and comments that appear on a line by themselves
+        # All trailing comments
+        # NOTE not including comments that appear on a line by themselves
```
gwelymernans, line 282: Please reflow this comment. And place any "NOTE" comments at the end of the comment block.

yapf/yapflib/reformatter.py
```diff
@@ -309,7 +311,16 @@ def _AlignTrailingComments(final_lines):
           line_content = ''
           pc_line_lengths = []
 
+          #NOTE
```
gwelymernans, line 314: Empty note.

yapf/yapflib/reformatter.py
```diff
@@ -309,7 +311,16 @@ def _AlignTrailingComments(final_lines):
...
+          contain_object = False
           for line_tok in this_line.tokens:
+
+            #NOTE if a line with inline comment is itself
```
gwelymernans, line 318: I don't think you need a "NOTE" tag here (or in some other places).

yapf/yapflib/reformatter.py
```diff
@@ -359,19 +376,31 @@ def _AlignTrailingComments(final_lines):
...
-              pc_line_length_index += 1
 
-              line_content = []
 
-              for comment_line_index, comment_line in enumerate(
-                  line_tok.value.split('\n')):
-                line_content.append('{}{}'.format(whitespace,
+              ''' this is added when we don't want comments on newlines
+                  to align with comments inline
+              '''
```
gwelymernans, line 383: This should be a comment instead of multiline string.

yapftests/yapf_test.py
```diff
@@ -26,7 +26,7 @@
 
 from lib2to3.pgen2 import tokenize
 
-from yapf.yapflib import errors
+from yapf.yapflib import errors, reformatter
```
gwelymernans, line 29: I don't think you need this.

gwelymernans (commented):
Thanks for the PR! Please run yapf over the changes. Also please update the CHANGELOG and README.rst files.

yapf/yapflib/format_decision_state.py
lizawang: Thanks for the replies. I will fix all the problems today.  We know that a dictionary, a list and an argument list is one logical line no matter how long it is with its entries all output in separate lines. Here it means that only when the comment is not inside this kind of logical line, aka, the comment has parent level equal to 0, we set the spaces_required_before to 0. Because we don't want the comments inside a dictionary to always align with its parent (the dictionary variable name). Examples are given in pull request README.

PR #1023, opened 2022-08-24
Fix fallback behavior on invalid/empty line ranges.

msuozzo:
The existing logic behaves poorly when the output of _LineRangesToSet is empty (note: not `None`, an empty set). Instead of restricting formatting to the empty set of lines (i.e. suppressing all formats), the current behavior lifts all line constraints and formats the whole file. It's perhaps not a bug but I think it's surprising behavior and the improper response to unexpected input.

### DIFF
--- a/yapf/yapflib/yapf_api.py
+++ b/yapf/yapflib/yapf_api.py
@@ -275,7 +275,7 @@ def _LineRangesToSet(line_ranges):
 
 def _MarkLinesToFormat(llines, lines):
   """Skip sections of code that we shouldn't reformat."""
-  if lines:
+  if lines is not None:
     for uwline in llines:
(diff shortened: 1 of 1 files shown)


gwelymernans:
Do you have a testcase for this?

msuozzo:
Done

thiagowfx:
> Done

Where?

bwendling:
This technically should never happen because `_LineRangesToSet` returns a non-null set if and only if there are line ranges specified.

msuozzo:
> > Done
> 
> Where?

oops! it never got pushed.

Regardless, I'm fine abandoning this given it's no longer relevant for our use-case.

PR #1026, opened 2022-09-23
Also support Black's `# fmt: off/on` pragma.

yilei:


gwelymernans:
Thanks! Could you add a note in the CHANGELOG?

yilei:
Done

PR #1040, opened 2022-11-30
Prevent crashes against valid 'pyproject.toml'.

emcd:
* Replace 'toml' dependency with 'tomli', which fully supports the current TOML specification.

The `toml` library does not fully support the current [TOML specification](https://toml.io/en/v1.0.0)... in particular, [arrays with items of heterogeneous types](https://toml.io/en/v1.0.0#array). When `toml` encounters things that it does not support, during load, it raises an exception which causes Yapf to crash in turn. Other tools in `pyproject.toml` can handle (and even expect) some of the things on which `toml` crashes.

The `tomli` library is noted as the inspiration for the `tomllib` module in Python 3.11+ standard library, according to [PEP 680](https://peps.python.org/pep-0680/#rationale). Furthermore, the author of `tomli` (who is one of the PEP authors) is actively maintaining `tomli` for backward compatibility at least until Python 3.10 reaches end-of-life, according to a [statement in the `tomli` repository](https://github.com/hukkin/tomli#intro):
> Tomli continues to provide a backport on PyPI for Python versions where the standard library module is not available and that have not yet reached their end-of-life.
Given this, `tomli` seems like the ideal replacement for `toml`.

Where the package is actually being used in the code, I have elected to use `import tomli as tomllib` to make it easier to switch to `tomllib` once Python 3.11 is the baseline supported version. In cases where the import is just being used a feature test, I simply added an "i" to the existing import with the belief that the feature tests can be removed once Python 3.11 is baseline.

In accordance with `HACKING.rst`, I ran:
* `PYTHONPATH=$PWD/yapf python3 -m yapf -i -r .`
* `python3 setup.py test`

before submitting this PR. Also verified directly that the command still runs correctly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,6 +81,6 @@ with codecs.open('README.rst', 'r', 'utf-8') as fd:
           'test': RunTests,
       },
       extras_require={
-          'pyproject': ['toml'],
+          'pyproject': ['tomli'],
       },
   )
--- a/yapf/yapflib/file_resources.py
+++ b/yapf/yapflib/file_resources.py
@@ -49,14 +49,15 @@ def _GetExcludePatternsFromPyprojectToml(filename):
   """Get a list of file patterns to ignore from pyproject.toml."""
   ignore_patterns = []
   try:
-    import toml
+    import tomli as tomllib
   except ImportError:
     raise errors.YapfError(
-        "toml package is needed for using pyproject.toml as a "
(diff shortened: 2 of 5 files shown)


emcd:
I use my Github cloaked emails for commits but the CLA agreement manager detected my Gmail address. Will look into registering my cloaked email tomorrow.

bwendling:
Thanks!

PR #1044, opened 2022-12-26
Generated dicts also have keys and values for rules to apply

alexey-pelykh:
The general idea:

If
```python
responseDict = {'childDict': {'spam': 'eggs'}}
```
is formatted as 
```python
responseDict = {
    'childDict': {
        'spam': 'eggs'
    }
}
```
so should
```python
generatedDict = {x: x for x in 'value'}
```
be formatted as 
```python
generatedDict = {
    x: x for x in 'value'
}
```
assuming `force_multiline_dict` is enabled. Also allows to achieve the following format if `split_before_dict_set_generator` is enabled:
```python
generatedDict = {
    x: x
    for x in 'value'
}
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 ### Changes
 - Moved 'pytree' parsing tools into its own subdirectory.
 - Add support for Python 3.10.
+- Format generated dicts with respect to same rules as regular dicts
 ### Fixed
 - Split line before all comparison operators.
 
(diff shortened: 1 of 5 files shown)


bwendling:
Thanks! Please add a comment to CHANGELOG.

alexey-pelykh:
@bwendling not sure that done correctly, yet done nonetheless

PR #1047, opened 2023-01-04
Deleted unnecessary symbol and conditional

thesayfulla:


hartwork (changes requested):
Hi @thesayfulla,

no objections to the simplification of the conditional but the CLA will need to be signed and dropping the round braces will give a syntax error in all known versions of Python. How would you like to move forward?

Best, Sebastian

thesayfulla:
Hello @hartwork 

I'm really sorry! My mistake.

The **IsCommentStatement** function should be like that:
```python
def IsCommentStatement(node):
  return NodeName(node) == 'simple_stmt' and node.children[0].type == token.COMMENT
```
or previous version is also good:
```python
def IsCommentStatement(node):
  return (NodeName(node) == 'simple_stmt' and
          node.children[0].type == token.COMMENT)
```

hartwork:
@thesayfulla cool that we're on the same page with this. For mergability, I believe we would need (1) the `IsCommentStatement` part reverted and (2) the CLA signed. (Please note that I cannot make any related promises because I have no merge permissions here myself.)

PR #1048, opened 2023-01-07
added knob count to README

SXHRYU:
I think it would be best to add overall number of knobs. I tried to play and tweak with all of them, but I wasn't sure if I had all of them in my **pyproject.toml** file. So I think there are *some* people who'd like to also have an exact number of available knobs. It doesn't clutter the README that much and it's easy to update the number in case of modifying the number of knobs.

P. S. I counted them using `print(len(_STYLE_HELP))` in `yapf/yapflib/style.py`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -390,6 +390,8 @@ Options::
 Knobs
 =====
 
+There are 60 knobs in total.
+
 ``ALIGN_CLOSING_BRACKET_WITH_VISUAL_INDENT``
     Align closing bracket with visual indentation.
 


hartwork (commented):
Hi @SXHRYU,

if this number is added to the docs then (a) it will go out of sync and become a mis-service or (b) it also needs additional CI to enforce that it stays in sync.

Personally, I would argue that when going through all options, the code is source number one regarding completeness and then arguably the content of `_STYLE_HELP.keys()` would be more helpful than just the count.

I think it's fair to say that in its current form the pull request is not a sustainable net positive, and it would need signing the CLA to be merged also.  How would you like to move forward?

Best, Sebastian

bwendling:
I agree with @hartwork. It's too likely to drift away from the "true" knob count. The `--style-help` option is probably the best for people anyway as it gives both the knobs and documentation.

hartwork:
@bwendling my vote for closing then.

PR #1054, opened 2023-02-07
Add --print-modified to print modified files

eltoder:
Prints modified files when in place mode is used. This helps debugging formatting issues and makes developers more aware of the changes made by yapf.

### DIFF
--- a/yapftests/yapf_test.py
+++ b/yapftests/yapf_test.py
@@ -505,6 +505,16 @@ class CommandLineTest(unittest.TestCase):
         reformatted_code = fd.read()
     self.assertEqual(reformatted_code, expected_formatted_code)
 
+  def testPrintModified(self):
+    for unformatted_code, has_change in [('1==2', True), ('1 == 2', False)]:
+      with utils.TempFileContents(
+          self.test_tmpdir, unformatted_code, suffix='.py') as filepath:
+        output = subprocess.check_output(
+            YAPF_BINARY + ['--in-place', '--print-modified', filepath],
+            text=True)
+        check = self.assertIn if has_change else self.assertNotIn
+        check(f'Formatted {filepath}', output)
+
   def testReadFromStdin(self):
     unformatted_code = textwrap.dedent("""\
         def foo():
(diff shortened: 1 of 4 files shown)


hartwork (approved):

bwendling:
Please add a note in the `CHANGELOG` and some tests please.

hartwork (changes requested):
Formally marking as "changes requested" to represent the comments by @bwendling above

eltoder:
@bwendling PTAL

yapftests/yapf_test.py
```diff
@@ -505,6 +505,16 @@ def testInPlaceReformattingEmpty(self):
...
 
+  def testPrintModified(self):
+    for unformatted_code, has_change in [('1==2', True), ('1 == 2', False)]:
+      with utils.TempFileContents(
+          self.test_tmpdir, unformatted_code, suffix='.py') as filepath:
+        output = subprocess.check_output(
+            YAPF_BINARY + ['--in-place', '--print-modified', filepath],
+            text=True)
+        check = self.assertIn if has_change else self.assertNotIn
+        check('Formatted %s' % filepath, output)
```
hartwork, line 516: Given that Python support for Yapf is at Python >=3.7 by now, how about:
```diff
--- check('Formatted %s' % filepath, output)
+++ check(f'Formatted {filepath}', output)
```
One more cases like that further up. Just an idea.

hartwork (commented):
@eltoder thanks for the updates! :+1: 

One idea for a small change below, not critical by any means:

hartwork:
> @bwendling PTAL

It's "Please Take A Look" in case you'd have to look it up like I did…

yapftests/yapf_test.py
eltoder: Done.

eltoder:
@hartwork Good point. We can use f-strings now. Done.

hartwork (approved):
@eltoder thanks for the adjustments and for being open to the idea! :+1:

bwendling:
Thank you for the PR and apologies for the late response.

PR #1061, opened 2023-02-24
ci: run on every PR and on main branch

thiagowfx:
Furthermore, update actions to the latest stable versions.


thiagowfx:
@bwendling I'll go ahead with this because it will make it easier to review the open PRs and ensure they don't break any tests, but if you have any concerns let me know and I can fix them in a follow-up.

Issue #1045, opened 2022-12-28
match case statement raise lib2to3.pgen2.parse.ParseError: bad input. Using yapf 0.32.0


tomina-s:
Hi.

Using Python3.10 match case raise exception.
Version: yapf-0.32.0

```
def http_error(status):
    match status:
        case _:
            return ""
```

```
Traceback (most recent call last):
  File "C:\Program Files\Python310\lib\site-packages\yapf\yapflib\pytree_utils.py", line 119, in ParseCodeToTree
    tree = parser_driver.parse_string(code, debug=False)
  File "C:\Program Files\Python310\lib\lib2to3\pgen2\driver.py", line 103, in parse_string
    return self.parse_tokens(tokens, debug)
  File "C:\Program Files\Python310\lib\lib2to3\pgen2\driver.py", line 71, in parse_tokens
    if p.addtoken(type, value, (prefix, start)):
  File "C:\Program Files\Python310\lib\lib2to3\pgen2\parse.py", line 162, in addtoken
    raise ParseError("bad input", type, value, context)
lib2to3.pgen2.parse.ParseError: bad input: type=1, value='status', context=(' ', (2, 10))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Program Files\Python310\lib\site-packages\yapf\yapflib\yapf_api.py", line 183, in FormatCode
    tree = pytree_utils.ParseCodeToTree(unformatted_source)
  File "C:\Program Files\Python310\lib\site-packages\yapf\yapflib\pytree_utils.py", line 125, in ParseCodeToTree
    tree = parser_driver.parse_string(code, debug=False)
  File "C:\Program Files\Python310\lib\lib2to3\pgen2\driver.py", line 103, in parse_string
    return self.parse_tokens(tokens, debug)
  File "C:\Program Files\Python310\lib\lib2to3\pgen2\driver.py", line 71, in parse_tokens
    if p.addtoken(type, value, (prefix, start)):
  File "C:\Program Files\Python310\lib\lib2to3\pgen2\parse.py", line 162, in addtoken
    raise ParseError("bad input", type, value, context)
lib2to3.pgen2.parse.ParseError: bad input: type=1, value='status', context=(' ', (2, 10))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Program Files\Python310\lib\site-packages\yapf\__init__.py", line 225, in _FormatFile
[...]

PR #1067, opened 2023-04-08
ylib2to3 -- Fork lib2to3 with some blib2to3 backports; add structured pattern matching (match) and parenthesized-context-managers statement support

### LINKED ISSUE #983: match case statement raise lib2to3.pgen2.parse.ParseError: bad input

Spitfire1900:
closes #983, #1045, #1085 (match)

Opened after @char101 said it was okay to open a PR for this. Many thanks to them for adding match statement support for yapf!

lib2to3 is no longer maintained, this forks lib2to3 and includes some modifications backported from black's blib2to3.

[I've looked into ](https://github.com/Spitfire1900/yapf3.10/tree/blib2to3) modifying yapf to work directly with blib2to3 instead of forking it but this work is in-progress and may not be possible. In addition it holds the risk that the black team does not consider blib2to3 as meant for other projects to consume.

I personally feel that a better long term approach would be for the black and yapf projects to externalize blib2to3 as a separate package that each project depends on rather than having each maintain their own fork.


Spitfire1900:
PR is for https://github.com/google/yapf/issues/1045

char101:
Although there are a lot of changes, the change to yapf itself is actually trivial. Here is what I did:

1. Copy lib2to3 from python 3.10 to yapf/yalib2to3
2. Manually apply the changes from the commits of https://github.com/psf/black/commits/main/src/blib2to3 which are not black specific
3. Change the imports from lib2to3 to ylib2to3
4. Add match and case block node to yapf

Spitfire1900:
We'll still need to fix some GitHub action errors and pytest failures before merging.

Meanwhile I'll get the CLA signed.

setup.py
```diff
@@ -51,6 +51,7 @@ def run(self):
       maintainer='Bill Wendling',
       maintainer_email='morbo@google.com',
       packages=find_packages('.'),
+      package_data={'': ['*.txt']},
```
thiagowfx, line 54: What is this for?

thiagowfx:
Could you add a few tests?

bwendling:
Thank you for this work! I need to consult internally to make sure we can allow this code into the tree.

In the meantime, there are some failures in the checks. Could you look into them?

setup.py
Spitfire1900: removed.

README.rst
```diff
@@ -15,6 +15,9 @@ YAPF
     :alt: Coverage status
 
 
+**YAPF fork with support for match/case statement (based on [blib2to3](https://github.com/psf/black/commits/main/src/blib2to3) and other personal
```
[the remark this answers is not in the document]
Spitfire1900: removed.

setup.py
char101: That's for Grammar.txt in lib2to3.

Spitfire1900:
I am currently at 8 failures with pytest on 3.10


```python
================================================================================================================================================= short test summary info =================================================================================================================================================
FAILED yapftests/reformatter_pep8_test.py::TestsForPEP8Style::testHangingIndentCollision - AssertionError: Code format mismatch:
FAILED yapftests/reformatter_python3_test.py::TestsForPython3Code::testContinuationIndentWithAsync - AssertionError: Code format mismatch:
FAILED yapftests/yapf_test.py::CommandLineTest::testReformattingSkippingLines - AssertionError: "def [65 chars]aaaa'\n        and xxxxxxxxxxxx.yyyyyyyy(zzzzz[249 chars]le\n" != "def [65 chars]aaaa' and\n            xxxxxxxxxxxx.yyyyyyyy(z[253 chars]le\n"
FAILED yapftests/yapf_test.py::CommandLineTest::testReformattingSkippingSingleLine - AssertionError: "def [65 chars]aaaa'\n        and xxxxxxxxxxxx.yyyyyyyy(zzzzz[233 chars]ss\n" != "def [65 chars]aaaa' and\n            xxxxxxxxxxxx.yyyyyyyy(z[237 chars]ss\n"
FAILED yapftests/yapf_test.py::CommandLineTest::testReformattingSkippingToEndOfFile - AssertionError: "def [65 chars]aaaa'\n        and xxxxxxxxxxxx.yyyyyyyy(zzzzz[461 chars]ss\n" != "def [65 chars]aaaa' and\n            xxxxxxxxxxxx.yyyyyyyy(z[465 chars]ss\n"
FAILED yapftests/yapf_test.py::CommandLineTest::testReformattingSpecificLines - AssertionError: "def [65 chars]aaaa'\n        and xxxxxxxxxxxx.yyyyyyyy(zzzzz[216 chars]ss\n" != "def [65 chars]aaaa' and\n            xxxxxxxxxxxx.yyyyyyyy(z[220 chars]ss\n"
FAILED yapftests/yapf_test.py::CommandLineTest::testRetainingHorizontalWhitespace - AssertionError: "def [65 chars]aaaa'\n        and xxxxxxxxxxxx.yyyyyyyy(zzzzz[250 chars]ss\n" != "def [65 chars]aaaa' and\n            xxxxxxxxxxxx.yyyyyyyy(z[254 chars]ss\n"
FAILED yapftests/yapf_test.py::CommandLineTest::testRetainingVerticalWhitespace - AssertionError: "def [65 chars]aaaa'\n        and xxxxxxxxxxxx.yyyyyyyy(zzzzz[220 chars]ss\n" != "def [65 chars]aaaa' and\n            xxxxxxxxxxxx.yyyyyyyy(z[224 chars]ss\n"
============================================================================================================================================= 8 failed, 591 passed in 21.78s ==============================================================================================================================================
```

Spitfire1900:
@char101 it looks like there is favorability with eventually merging in this PR but there are some failing tests. Resolving these are a little bit out of my wheelhouse, would you be interested in submitting changes to get [the above mentioned tests](https://github.com/google/yapf/pull/1067#issuecomment-1522152182) passing?

char101:
Maybe you could try reverting https://github.com/google/yapf/pull/1067/commits/586c299c74c6c6458743bb0fa3f4432c678c9e6b since that is just my personal modification.

Spitfire1900:
@char101 are you able to sign the cla?
[cla/google Failing after 2s — ❌ Missing CLA from one or more contributors ](https://github.com/google/yapf/pull/1067/checks?check_run_id=13160766983)

Spitfire1900:
So open question, how closely should ylib2to3 mirror blib2to3? I elected at this time to change it as necessary to pass flake8 tests where I felt reasonable.

char101:
> @char101 are you able to sign the cla? [cla/google Failing after 2s — ❌ Missing CLA from one or more contributors ](https://github.com/google/yapf/pull/1067/checks?check_run_id=13160766983)

Done

> So open question, how closely should ylib2to3 mirror blib2to3? I elected at this time to change it as necessary to pass flake8 tests where I felt reasonable.

Changes to blib2to3 (mostly Grammar.txt nowadays, changes to the code are mostly black related) need to be applied by hand so there is no need to mirror those.

Spitfire1900:
Tests are passing and this is ready for review! Many thanks to @char101 for doing all the hard work :wink:

char101:
Other things that might be missing/to consider

1. Add credit to black/blib2to3
2. blib2to3 (or in fact the whole black itself) is now compiled to native extensions using mypyc, so it might worth to consider using mypyc too, maybe starting with ylib2to3 since yapf

Spitfire1900:
@char101 
1. I've added credit to black in a nested README in the ylib2to3 directory, I also added a comment in its __init__.py
2. Regarding mypyc, I would consider that a .NEXT enhancement outside the scope of this PR.

Spitfire1900:
CC @hartwork and @bwendling

bwendling:
Okay, I just got approval to accept third-party code. It'll need to be adjusted though. I moved the `yapf/yapf/third_party/` sub-directory to `yapf/third_paty`. You'll have to place the code into that directory. It also needs the `LICENSE` file from the project it's copying.

Spitfire1900:
I did so but I have some concerns

1. Is throwing this into a `third_party` package really the best choice? `yapf.third_party` seems better.
2. I am running into this error when performing test uninstalls on my system. Does anyone have a suggestion to remove this warning and remove the *.pickle files?

```
$ pip uninstall yapf
Found existing installation: yapf 0.33.0
Uninstalling yapf-0.33.0:
  Would remove:
    /home/kyle/Repos/py_sandbox/.venv/bin/yapf
    /home/kyle/Repos/py_sandbox/.venv/bin/yapf-diff
    /home/kyle/Repos/py_sandbox/.venv/lib/python3.10/site-packages/third_party/*
    /home/kyle/Repos/py_sandbox/.venv/lib/python3.10/site-packages/yapf-0.33.0.dist-info/*
    /home/kyle/Repos/py_sandbox/.venv/lib/python3.10/site-packages/yapf/*
    /home/kyle/Repos/py_sandbox/.venv/lib/python3.10/site-packages/yapftests/*
  Would not remove (might be manually added):
    /home/kyle/Repos/py_sandbox/.venv/lib/python3.10/site-packages/third_party/ylib2to3/Grammar3.10.6.final.0.pickle
    /home/kyle/Repos/py_sandbox/.venv/lib/python3.10/site-packages/third_party/ylib2to3/PatternGrammar3.10.6.final.0.pickle
Proceed (Y/n)? 
```

char101:
third_party is C directory structure, it doesn't make sense in a python project. The way to make a third paty in a python project is to make it a separate package outside yapf (just like blib2to3 which have its own package). But the question is, is it even considered a third party code. A third party code should be a separate library maintained by other people.

Spitfire1900:
[0e86a24](https://github.com/google/yapf/pull/1067/commits/0e86a24c1d7ccb260301b5c371b1b0e4412b863f) might be terrible jank but it solves the 'pickle' problem.

ylib2to3's pickle load presumably breaks yapf running in a PyZipFile.

Spitfire1900:
Made an attempt to fix the PyZipFile case, but still getting this error. Unsure how to resolve.

```python
$ python -m yapf
  File "/usr/lib/python3.10/runpy.py", line 187, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/lib/python3.10/runpy.py", line 146, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "/usr/lib/python3.10/runpy.py", line 110, in _get_module_details
    __import__(pkg_name)
  File "/home/kyle/Repos/py_sandbox/dependencies.zip/yapf/__init__.py", line 39, in <module>
  File "/home/kyle/Repos/py_sandbox/dependencies.zip/yapf/yapflib/yapf_api.py", line 40, in <module>
  File "/home/kyle/Repos/py_sandbox/dependencies.zip/yapf/pytree/pytree_unwrapper.py", line 34, in <module>
  File "/home/kyle/Repos/py_sandbox/dependencies.zip/yapf/pytree/pytree_utils.py", line 30, in <module>
  File "/home/kyle/Repos/py_sandbox/dependencies.zip/third_party/ylib2to3/pygram.py", line 29, in <module>
  File "/home/kyle/Repos/py_sandbox/dependencies.zip/third_party/ylib2to3/pgen2/driver.py", line 247, in load_packaged_grammar
  File "/home/kyle/Repos/py_sandbox/dependencies.zip/third_party/ylib2to3/pgen2/grammar.py", line 100, in loads
_pickle.UnpicklingError: invalid load key, '#'
```

char101:
Can't the pickle files be added to `package_data` like `Grammar.txt`? `package_data` supports glob so you can just add `*.pickle` there.

Spitfire1900:
[96ec816 -- add compiled pickles to package output](https://github.com/google/yapf/pull/1067/commits/96ec816f0ad3fe8f2a98c1194a59012d6a896b39) resolves this, but now my concern is that the release artifact must be built with python3.7. My understanding is thatbuilding on 3.8 or newer will cause the pickle files to build using version 5 instead of version 4.

https://docs.python.org/3/library/pickle.html#data-stream-format

I have manually checked on my local machine that:
1. pickle files are generated when `import yapf` runs during `python -m setup bdist_wheel` so they are available by the time the package is being assembled
1. I can run yapf as a pyzipfile
1. Installation and uninstallation does not produce errors.

char101:
I think ideally the pickle could be generated at install time (using `install` `cmdclass` in `setup.py`) by running `pgen.Grammar.dump`.

https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools

char101:
What blib2to3 does is to generate the pickle in the package installation directory and just ignore if the write failed.

https://github.com/psf/black/blob/eb32729ab562c010d9f644a3bbb2305196b2deb7/src/blib2to3/pgen2/driver.py#L270

Since the user is more likely to install yapf in the user python path or in a virtual env, I think this is also a viable approach (i.e. try to generate the pickle when running yapf but just ignore if writing the pickle file failed).

Spitfire1900:
I will look into that. I am also willing to accept PRs on my version of the feature branch.

char101:
Just to add timing data on the effect of the pickle cache

```python
In [10]: %timeit load_grammar(save=False)
14.9 ms ± 352 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [11]: %timeit load_grammar(save=True)
399 µs ± 17.6 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

I don't think 15ms is that significant so we can pick the simplest solution i.e. to ignore writing the pickle if the ylib2to3 directory is not writable.

Spitfire1900:
@char101 @bwendling @hartwork Ultimately regarding pickle files I decided the following.
1. That it was safer to have an unclean uninstall (pickle files are left after pip uninstall) then to implement a cache dir at this time
2. That pickle files should not be included in the package because you have to make sure they are generated as version 4 for python3.7
3. That pickle files should not be generated as a post-install action since that capability only works with sdist and is going away soon.

So:
1. Grammer files are pickled into `site_packages` on demand if the grammer file path is valid.
2. The Grammer file's content is streamed to third_party.ylib2to3.pgen2.generate_grammar() as a StringIO object and not cache otherwise.

Spitfire1900:
1. Are we talking site_packages as in my local /usr/lib/python3.11/site-packages/ where only root can write? Are we talking generation at runtime (where permissions are lacking) or at distro package creation?
A. The cached pickle file is attempted to be created at runtime. The cache creation will only work when site-packages is writable by the current user. When installed as root the associated OSError is caught at the pickle is loaded into memory instead. https://github.com/Spitfire1900/yapf/blob/3.10/third_party/ylib2to3/pgen2/driver.py#L213-L224

2. I will work on getting ylib2to3 and yapf_diff moved to `yapf.third_party`

hartwork:
@Spitfire1900 thanks for your reply and for being up for resolving use of system-wide package `third_party` :+1: 
Regarding the grammar pickle cache, I like that the Python version is part of the name. It would be great to also include the version of yapf so that on yapf updates users are guaranteed not working with an out-of-sync grammar cache (unless impossible by definition?). For cache data location I would like to suggest `$XDG_CACHE_HOME` for something that will feel at home with regard to the [Filesystem Hierarchy Standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) and the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). What do you think?

Spitfire1900:
CC @hartwork on the following:
> Cache behavior changes from the currently proposed may be better suited in a follow-up PR rather than implemented in the current branch.

Spitfire1900:
@hartwork and @char101 is it okay if this PR is merged in and cache behavior improvements are added in a follow-up PR or do those improvements need to be part of this PR?

What other outstanding concerns are there?

Spitfire1900:
parenthesized-context-managers support is also brought in by this change (just realized, thanks @char101).

hartwork:
Hello @Spitfire1900,


what I did now is this:
- clone #1067 locally
- clone black locally
- auto-format both *lib2to3 libraries using the same formatter (black 23.3.0 in my case)
- _superficially_ inspect the resulting diff using [Meld](https://meldmerge.org/)

For anyone to reproduce on the terminal:
```bash
git clone https://github.com/Spitfire1900/yapf
git clone https://github.com/psf/black
black black/src/blib2to3/ yapf/yapf/third_party/ylib2to3/
meld black/src/blib2to3/ yapf/yapf/third_party/ylib2to3/ &
```

Before I go into details, I would like to note more general that:

- I have no power or permissions in yapf, just thoughts.
- After auto-formatting, I expected to find a minimal diff, but the diff contains "noise", in particular type annotation are dropped/missing, e.g. `def _newer(a: Text, b: Text) -> bool:` in *b*lib2to3 is `def _newer(a, b):` in *y*lib2to3. If the plan is maintain a living fork with frequent porting of changes from *b*lib2to3 to *y*lib2to3, that part of the diff may cause a lot of additional unnecessary(?) friction. Maybe black made these changes in the mean time, I did not check.
- There is no (high level or detailed) documentation of the changes done to get from *b*lib2to3 to *y*lib2to3 so this will have a tiny [bus factor](https://en.wikipedia.org/wiki/Bus_factor) and raise the question if this will be maintainable and by/for who.
- I'm noticing only now that ylib2to3 with changes is a lot more *first* party than third.
- I wonder if adding this code extends the public API of yapf.  If that is undesired, potentially some files/folders would need to be renamed to start with an underscore.

Regarding two details that caught my eye:
- The *b*lib2to3 readme seems to have value that is not carried over
- Method `load_packaged_grammar` at `yapf/third_party/ylib2to3/pgen2/driver.py` is ~90% dead code because of the return statement in the first line. I'm unsure about the intended behavior, but the current code is unlikely to be intended as is.

char101:
blib2to3 is black internal library and it contains many blank internal modifications.

As I have written above, I wasn't copying blib2to3 into yapf, I was copying lib2to3 into yapf, using the changes made into blib2to3. And that modification into lib2to3 is made to minimize changes, which is why type annotations wasn't there.

Spitfire1900:
@char101 Do you know from which black commit ylib2to3 was originally forked from and from which it was last synced?

What was the major blockers in having a clean fork over a fairly modified one?

Spitfire1900:
- After auto-formatting, I expected to find a minimal diff, but the diff contains "noise", in particular type annotation are dropped/missing, e.g. def _newer(a: Text, b: Text) -> bool: in blib2to3 is def _newer(a, b): in ylib2to3. If the plan is maintain a living fork with frequent porting of changes from blib2to3 to ylib2to3, that part of the diff may cause a lot of additional unnecessary(?) friction. Maybe black made these changes in the mean time, I did not check.
   - Project is lib2to3 with some backporting of blib2to3 features [per @char101](https://github.com/google/yapf/pull/1067#issuecomment-1544422427)
- There is no (high level or detailed) documentation of the changes done to get from blib2to3 to ylib2to3 so this will have a tiny [bus factor](https://en.wikipedia.org/wiki/Bus_factor) and raise the question if this will be maintainable and by/for who.
    - I think this is effectively _unavoidable_ to accept the low bus factor in order for yapf to stay relevant.
- I'm noticing only now that ylib2to3 with changes is a lot more first party than third.
   - I would still consider this primarily third_party ATM.
- I wonder if adding this code extends the public API of yapf. If that is undesired, potentially some files/folders would need to be renamed to start with an underscore.
   - yes, `from yapf.third_party import ylib2to3` works

char101:
blib2to3 is tightly related with black, so rather than thinking about syncing blib2to3, I think it is easier to think about fixing lib2to3 (which is the actually the initial problem) by borrowing code from blib2to3.

Other than some changes to support context sensitive keyword (match and case), there is actually little to no changes required (in the future) to lib2to3 python codes, probably only come modifications to Grammar.txt to support recent PEPs as can be seen in blib2to3 commit history.

Spitfire1900:
@char101 I do agree with @hartwork that `yapf/third_party/_ylib2to3/README.rst` requires a changeset of what backported to it from blib2to3

bwendling:
@char101 
> blib2to3 is tightly related with black, so rather than thinking about syncing blib2to3, I think it is easier to think about fixing lib2to3 (which is the actually the initial problem) by borrowing code from blib2to3.

Could you explain some of the things that make `blib2to3` tightly related to Black? The original `lib2to3` is a (mostly) generic parser. If `blib2to3` generates the same or similar ptree structure then it should be usable for us.

bwendling:
@Spitfire1900 

> * Is throwing this into a `third_party` package really the best choice? `yapf.third_party` seems better.

Let me recheck the requirements.

Spitfire1900:
>Could you explain some of the things that make blib2to3 tightly related to Black? The original lib2to3 is a (mostly) generic parser. If blib2to3 generates the same or similar ptree structure then it should be usable for us.

In my basic testing I have found that blib2to3 breaks the lib2to3 API.

https://github.com/Spitfire1900/yapf/commits/blib2to3 is my attempt to reference blib2to3 directly. My biggest blocker right now is this:
```python
>       parser_driver = driver.Driver(_GRAMMAR_FOR_PY3, convert=pytree.convert)
E       TypeError: 'convert' is an invalid keyword argument for __init__()
```

Spitfire1900:
Now, as I mention in the opening summary for this PR the long term optimal outcome is that yapf use blib2to3 instead of ylib2to3. But I think since we've gotten the yapftest suite to pass on ylib2to3 today that migrating to blib2to3 can be a future endeavor. 

Also, migrating to blib2to3 requires assurances from the black team that blib2to3's api is stable and not going to break between black releases, and/or that it be extracted to its own pypi entry and depended on by both projects.

Spitfire1900:
While not in a private module blib2to3 is not considered public.

https://discord.com/channels/267624335836053506/846434317021741086/1106588410040303679

hartwork:
> @hartwork
> 
> > * Method load_packaged_grammar at yapf/third_party/ylib2to3/pgen2/driver.py is ~90% dead code because of the return statement in the first line. I'm unsure about the intended behavior, but the current code is unlikely to be intended as is.
> 
> This was intentional to not change function callpaths while also not attempting to load packaged grammars because of earlier mentioned concerns.

@Spitfire1900 I don't understand. Function `load_packaged_grammar` is not in a healthy state right now. Please help me understand. Thank you.

Spitfire1900:
@hartwork I've undone the changes to `load_packaged_grammar()` and have refactored any calls to it to simply call `load_grammar()` instead.

hartwork:
@Spitfire1900 thanks!

Spitfire1900:
@char101 can you open a PR against [Spitfire1900:3.10](https://github.com/Spitfire1900/yapf) for `yapf/third_party/_ylib2to3/README.rst` that includes which changes from blib2to3 were backported to ylib2to3?

char101:
No, that's what commit history is for.

Spitfire1900:
My best ability to identify the differences of ylib2to3 from lib2to3, I can add this to the readme:
- in pgen2.driver: ReleaseRange and TokenProxy classes
- in pgen2.parse: Recorder class
- soft_keywords
- some old syntax support for iterables, for loops, and if statements
- match statement support
- removal of `-m lib2to3` support
- removal of btm_utils and btm_matcher modules as well as the refactor module
- Updated to Grammar.txt to support Python3.10 features

This was done by
1. copying Python 3.10.7 source code for lib2to3 into `./yapf/third_party`
2. running `git add --all`
3. running `pre-commit run --all`
4. And then comparing in Meld

Spitfire1900:
I've looked at autopep8 and it too is still using lib2to3 https://github.com/search?q=repo%3Ahhatto%2Fautopep8%20lib2to3&type=code

Spitfire1900:
@hartwork @char101 @bwendling 

My general proposal
1. Review and merge in ylib2to3
2. Develop a PoC impl. of using blib2to3 instead of ylib2to3
3. Provided ^ is successful reach out to @JelleZijlstra about moving blib2to3 to a separate PyPi package
4. Extract blib2to3 to its own package and migrate yapf to it

Medium term following 4: black and yapf (and also possibly autopep8) migrate to the dedicated blib2to3 package.
Long-term: the three formatters move to using https://peps.python.org/pep-0617/

I'm going for small-as-possible changes here rather than large ones that affect multiple projects hence the suggestion for adopting ylib2to3 today and moving on blib2to3 later.

Spitfire1900:
> While not in a private module blib2to3 is not considered public.
> 
> 
> > Jelle#0575 it's not a public UI, and we may remove it if we come up with some other parser
> > I wouldn't be averse to making it into a separate installable project though, if there's interest
> > ~ https://discord.com/channels/267624335836053506/846434317021741086/1106588410040303679

Additional comments 

> though thinking more about this, it may cause complexities with packaging/mypyc compilation, so not sure if it would be worth the effort for us

Spitfire1900:
The following is not a great sign that blib2to3 is that simple to migrate to:

```
>     setattr(node, _NODE_ANNOTATION_PREFIX + annotation, value)
E     AttributeError: 'Node' object has no attribute '_yapf_annotation_child_indent'

yapf\pytree\pytree_utils.py:244: AttributeError
============================================== short test summary info ============================================== 
FAILED yapftests/blank_line_calculator_test.py::BasicBlankLineCalculatorTest::testDecorators - AttributeError: 'Node' 
object has no attribute '_yapf_annotation_child_indent'
========================================= 1 failed, 598 deselected in 0.47s =========================================
```

See https://github.com/Spitfire1900/yapf/tree/blib2to3

JelleZijlstra:
That's likely because blib2to3 is mypyc-compiled.

Spitfire1900:
@JelleZijlstra . Yep this sounds accurate. Installing black as an editable (non-mypyc) install had a lot more success.


```
=============================================================================== short test summary info ================================================================================
FAILED yapftests/comment_splicer_test.py::CommentSplicerTest::testCommentIsFirstChildInCompound - AssertionError: 'COMMENT' != 'INDENT'
FAILED yapftests/comment_splicer_test.py::CommentSplicerTest::testCommentsOnDedents - AssertionError: 'COMMENT' != 'INDENT'
FAILED yapftests/reformatter_buganizer_test.py::BuganizerFixes::testB18256826 - AssertionError: Code format mismatch:
FAILED yapftests/reformatter_buganizer_test.py::BuganizerFixes::testB30087362 - AssertionError: Code format mismatch:
FAILED yapftests/reformatter_buganizer_test.py::BuganizerFixes::testB30087363 - AssertionError: Code format mismatch:
FAILED yapftests/reformatter_facebook_test.py::TestsForFacebookStyle::testCommentWithNewlinesInPrefix - AssertionError: Code format mismatch:
FAILED yapftests/reformatter_python3_test.py::TestsForPython3Code::testAsyncFunctions - AssertionError: Code format mismatch:
FAILED yapftests/yapf_test.py::CommandLineTest::testDisableWithLinesOption - AssertionError: '# ya[62 chars]nner_func():\n        return\n    return\n    # yapf: enable\n' != '# ya[62 chars]nner_func():\n        return\n    return\n# yapf: enable\n'
FAILED yapftests/yapf_test.py::CommandLineTest::testReformattingSkippingLines - AssertionError: "def [310 chars]mmmmmmm[0]) == 'bbbbbbb'):\n        pass\n    # yapf: enable\n" != "def [310 chars]mmmmmmm[0]) == 'bbbbbbb'):\n        pass\n# yapf: enable\n"
FAILED yapftests/yapf_test.py::CommandLineTest::testRetainingVerticalWhitespace - AssertionError: 'if a[44 chars]n    same\n\n\n\n    #comment\n\n    #   trailing whitespace\n' != 'if a[44 chars]n    same\n\n\n\n#comment\n\n#   trailing whitespace\n'
=========================================================================== 10 failed, 589 passed in 29.64s ============================================================================
```

Spitfire1900:
@JelleZijlstra my observation is that all of the failing tests mentioned above include indented comments. Does blib2to3 handle this differently than lib2to3 intentionally?

JelleZijlstra:
Not that I know of.

Spitfire1900:
I have pushed a PoC implementation of using a Cache dir. `_generate_pickle_name()` uses `platformdirs.user_cache_dir()`.

On Windows this results in the following sample picklefile location:
`%LOCALAPPDATA%\Google\YAPF\Cache\0.33.0\Grammar-py3.10.7.final.0.pickle`

I still need to exposed the yapf version to the driver module without causing a circular reference or redefining the yapf version in multiple locations.

Spitfire1900:
All right, I’m running into a works on my machine problem between my workstation and GitHub CI.

But I have fixed the circular reference issue and version being specified twice. The ylib2to3 driver and setup.py reference data in new file ./yapf/PKG_INFO.ini

Spitfire1900:
Fixed by using using https://pypi.org/project/importlib-metadata

@hartwork now on Windows this results in the following sample picklefile location:
%LOCALAPPDATA%\Google\YAPF\Cache\0.33.0\Grammar-py3.10.7.final.0.pickle

hartwork:
> Fixed by using using https://pypi.org/project/importlib-metadata

@Spitfire1900 …and `platformdirs`! Cool! :+1:

Spitfire1900:
This PR is ready for review.

Spitfire1900:
Requesting review from @bwendling , @hartwork , and/or @eliben.

Spitfire1900:
Fixes walrus `:=` operator errors

```diff
> yapf --diff .\yapf#894.py
--- .\yapf#894.py       (original)
+++ .\yapf#894.py       (reformatted)
@@ -1,4 +1,5 @@
 import os
-a=[1,2,3,4]
-if (n:=len(a))>2:
+
+a = [1, 2, 3, 4]
+if (n := len(a)) > 2:
     print()
```

.gitignore
```diff
@@ -34,4 +35,5 @@
 /yapf.egg-info
 
 /.idea
+.vscode/settings.json
```
hartwork, line 38: Not important but here^ I would personally expect (a) `/.vscode/settings.json` with a trailing slash or just `/.vscode/` also with trialing slash here. Is status quo intended?

---

Apart from that, the parts of the pull request that I do understand look pretty good to me, this will be great gift to the yapf community :+1:

Spitfire1900: Used `/.vscode/settings.json`, excluding all of `/.vscode/` prevents adding things like launch.json, extensions.json.

hartwork: Thanks!

bwendling:
Thank you for all of your work! I really appreciate it. I was off last week for a family emergency so wasn't able to keep up. I'll review this as soon as I can.

bwendling:
One comment. The `third_party` directory needs to exist at the top-level of the code base. It's a requirement so that licenses are covered correctly. Could you please move it there and add a symlink to where it was so that things will still work?

hartwork:
> One comment. The `third_party` directory needs to exist at the top-level of the code base. It's a requirement so that licenses are covered correctly.

Hi @bwendling could you elaborate which entity or tool enforces this requirement?

Let me use the occasion to re-state that third party code with changes is _first_ party code, arguably. It's not too late to bypass third party rules by treating it as the first party code it de-facto becomes. Just my 2 cents.

> Could you please move it there and add a symlink to where it was so that things will still work?

I won't be surprised if the symlink breaks things… Is doing the linking the other way around, so that third_party is a symlink, instead? Just brainstorming if we need alternatives for technical reasons…

bwendling:
Google is the one enforcing this requirement (third-party code needs to be in a directory titled `third_party` at the top level). It's the provenance of the code that we need to document. If it's a (modified) copy of existing code, we need to retain that knowledge and document the license accordingly. (The corollary is we can't change the licensing either.)

> I won't be surprised if the symlink breaks things… Is doing the linking the other way around, so that third_party is a symlink, instead? Just brainstorming if we need alternatives for technical reasons…

That might be sufficient, since it's more about documentation rather than the structure.

Spitfire1900:
@bwendling is this a requirement for the Repo, or for the python site-packages structure? I may not be able to avoid "polluting" site packages with a `third_party` directory if it's moved to the root of this repo.

Are there other Google Python projects that do this already as a reference? The following search is coming up empty:
https://github.com/search?q=org%3Agoogle+language%3APython+path%3A%2Fthird_party&type=code

hartwork:
> @bwendling is this a requirement for the git repo, or for the python site-packages structure?

@Spitfire1900 that was explained above at https://github.com/google/yapf/pull/1067#issuecomment-1563600510 to be a Google thing. Does that answer the question you have in mind?

> I may not be able to avoid "polluting" site-packages with a `third_party` directory if it's moved to the root of this repo.

Adjusting the call at…

https://github.com/google/yapf/blob/88b72a25135230458006ca79726836bb7746375f/setup.py#L55

…could be the fix for that I believe.
A global package `third_party` in site-packages is not something I can support, it goes against sane PyPI practices, and that's already the nicest way I can put it :smiley:

Spitfire1900:
So https://github.com/Spitfire1900/yapf/tree/3.10-third-party is a mostly-successful example of fulfilling both of thes requirements:

1. Have a `third_party` directory at the root of the git repo
2. Do not install a `site-packages/third_party` directory at install time.

This:

1. Moves `yapf/third_party` content to `third_party/yapf_third_party` in the repo.
2. Installs `yapf_third_party` into `site-packages/yapf_third_party` at install time.

### _**BUT:**_ this _BREAKS_ editable installs (e.g. `pip install -e .`)**!!!**

Pip will add a `site-packages/easy-install.pth` with an entry for `<workspace>`. Installing the project as an editable install requires manually modifying this file to include an extra line for `<workspace>/third_party`.

Without a proper fix for this I'm going to have to suggest installing a `site-packages/third_party` directory at install time as the recommended approach, much to the chagrin of the larger Python community.

Spitfire1900:
Wait, so nevermind 🤦 . This is an issue with caching to some degree, editable installs actually do work... ARGH.

Spitfire1900:
Alright, requests changed made and merged in upstream.

Spitfire1900:
Finished up some GH workflow work as installing the project prior to running pytest is now required as dependencies `importlib-metadata` and `platformdirs` have been added.

Spitfire1900:
CC'ing @bwendling and @thiagowfx for review.

Spitfire1900:
What is preferred for site-packages layout for each of the third party modules? Should _ylib2to3 or yapf_diff be at the root of site-packages? Only yapf_diff? Neither? 

This is current:
```
site-packages
\_ yapf_third_party
   \_ _ylib2to3
   \_ yapf_diff
\_ yapf-0.33.0.dist-info
\_ yapf
\_ yapftests
```
I think that this is probably a better fit, or at least that yapf_diff should exist at the root of site-packages
```
site-packages
\_ _ylib2to3
\_ yapf_diff
\_ yapf-0.33.0.dist-info
\_ yapf
\_ yapftests
```

bwendling:
> What is preferred for site-packages layout for each of the third party modules? Should _ylib2to3 or yapf_diff be at the root of site-packages? Only yapf_diff? Neither?
> 
I don't have much of an opinion on this, but I think we should do that in a separate PR to not jar things too much.

Apologies for the lateness of this review. I'm dealing with emergency issues right now, so work has been spotty.

bwendling:
Thank you for taking this on! This is a major step forward in the life of YAPF.

Spitfire1900:
@bwendling Thank you very much, may this be released to PyPi?

thinkwelltwd:
@Spitfire1900, I've watched this PR with bated breath throughout its long development. Emoji reactions just aren't adequate to communicate my thanks and appreciation for your work.

And thank you, @bwendling, for yapf!  It's way and above the best python formatter, as far as I'm concerned.

Spitfire1900:
Most of the credit goes to @char101  for the adding the necessary logic to lib2to3 to support new language features. My work was primarily prepping it for merge into upstream. ☺️

bwendling:
I concur with @thinkwelltwd  :-)

This probably deserves its own PyPi release. I'll do one soon.

Again, thank you for all of your work, @Spitfire1900!

bwendling:
And thanks to @char101 for your work! This pushes YAPF past a major hurdle.

hartwork:
CC #1110

eli-schwartz:
> Fixed by using using https://pypi.org/project/importlib-metadata

I would always recommend you do a try/except and see if you can first use https://docs.python.org/3/library/importlib.metadata.html

This reduces the dependencies and makes it faster to install. It is available in the stdlib for nearly every supported python version. :)

Spitfire1900:
@eli-schwartz , this is a recommendation that has been advised as part of the tomllib PR: https://github.com/google/yapf/pull/1152

Spitfire1900:
Third party dependencies added by this PR are removed for newer python versions with #1158

PR #1069, opened 2023-04-12
Support pyproject.toml by default

Spitfire1900:


Spitfire1900:
@bwendling Do you think that we can publish a new release of yapf (0.33.0) for this?

PR #1073, opened 2023-04-18
Pin GitHub Actions to specific commits for security

hartwork:
For proof that GitHub Dependabot can still keep things up to date for us with the new format see https://github.com/gentoo-ev/www.gentoo-ev.org/pull/5/files .

(Also, this related interactive tool could be of interest: https://app.stepsecurity.io/secureworkflow/)
(And for additional context: https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,9 +16,9 @@ jobs:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
         os: [ubuntu-latest, macos-latest]
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab  # v3.5.2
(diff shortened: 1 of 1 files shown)


bwendling:
Is using the SHA1s the standard for this? If so, could you please note that in the description with a URL?

hartwork:
> Is using the SHA1s the standard for this? If so, could you please note that in the description with a URL?

It's the most secure flavor, but it's not as widely known as it should be. I have added more info to the description just now. What do you think?

hartwork:
@bwendling PS: I can add a GitHub Dependabot config to make it send pull requests to yapf either in here or in a dedicated pull request after, if you like.

hartwork:
@bwendling what do you think?

bwendling:
Sorry for the late response.

PR #1074, opened 2023-04-18
Drop Python <3.7 leftovers

hartwork:


yapf/yapflib/py3compat.py
```diff
@@ -18,9 +18,9 @@
 import os
 import sys
 
-PY3 = sys.version_info[0] >= 3
-PY36 = sys.version_info[0] >= 3 and sys.version_info[1] >= 6
-PY37 = sys.version_info[0] >= 3 and sys.version_info[1] >= 7
+PY3 = True
```
thiagowfx, line 21: Strictly speaking, we do not need any of these variables anymore.

Would you be up to removing them?

I'll approve this PR anyway because it moves the needle forward, but it would be nice if you sent a follow up to delete these variables.

hartwork: @thiagowfx the only reason I kept these symbols is because it could be argued that these symbols are part of the public API — they do not start with an underscore — so I didn't want to technically break public API without someone asking for it and being aware of these arguable consequences. I'll go ahead dropping them now…

hartwork: @thiagowfx most of `py3compat.py` is gone now, only three symbols are left:

- `PY38`
- `raw_input`
- `removeBOM`

I can move the last two to a different module, if you like. Happy about naming suggestions.

What do you think about the new status quo?

thiagowfx: Looks good, thanks!

bwendling:
I mentioned in another PR, but could you add comments in `CHANGELOG` and the `README` about dropping <3.7?

hartwork:
@bwendling okay, going for a dedicated pull request now :+1:

PR #1084, opened 2023-04-25
Remove more Python 2 code leftovers (tool 2to3)

hartwork:
Follow-up to #1074

CC @bwendling

### DIFF
--- a/yapf/__init__.py
+++ b/yapf/__init__.py
@@ -25,7 +25,6 @@ formatting decisions based on what's the best format for each line.
 
 If no filenames are specified, YAPF reads the code from stdin.
 """
-from __future__ import print_function
 
 import argparse
 import logging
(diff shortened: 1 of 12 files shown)


hartwork:
> After this, do we need the `yapflib/py3compat.py` file anymore?

@bwendling it still defines three symbols:

- PY38
- raw_input
- removeBOM

We can move things elsewhere, and then delete the file if you like
The first two symbols seem test-related, the third is used by the core at a single place.

To summarize the two key questions:
- Should the file be resolved by moving the related code?
- If yes, should it be done in here or in a follow-up pull request?

PS: Resolving conflicts in a second…

bwendling:
Okay, so I suggest moving `PY38` and `raw_input` to the `yapftest/` subdirectory and the placing the `removeBOM` into the file using it. That should be done as a separate PR.

Thank you!

hartwork:
@bwendling sounds good, PR upcoming…

Issue #1081, opened 2023-04-20
Enforce pre-commit clean code via CI?


hartwork:
Hi!

I'm considering to provide a pull request that enforces pre-commit clean CI because `main` is not pre-commit clean today and CI would prevent that from happening again. For that pull request to be healthy:
- pre-commit files would have ideally gone down to a single copy already (issue #1077) at that point
- Pull requests #1078, #1079, #1080 would ideally have been merged already
- It would need to be decided if (or when) mass-applying yapf to itself is considered tolerable with the current set of open pull requests around.

What do you think?

Spitfire1900:
That is fine, I would also like to see pre-commit handle virtual environment installs of the workspace. Pre-commit today works in terminals when a venv is activated but this does not work when committing via the Source Control pane of VSCode.

hartwork:
@Spitfire1900 I'm not entirely sure how that relates to CI — could you help me see the connection?

Regarding virtual environments, pre-commit can be used completely without virtualenvs, e.g. I have package `dev-vcs/pre-commit` installed system-wide on my Gentoo box. Many other distros [have it packaged](https://repology.org/project/python:pre-commit/badges) too.

Regarding making local commits, Git commit hooks as e.g. activated by `pre-commit install --install-hooks` should be running whenever someone makes new Git commits through Git as a subprocess unless passing `--no-verify`. So dependending on how IDEs put new commits into Git, it may just need activation of the related pre-commit hook.

PS: I have [six pull request waiting for review](https://github.com/google/yapf/pulls/hartwork) in here today, so I will postpone creating more until that number is near zero. If you can help with review and/or merging, that would rock.

What do you think?

Spitfire1900:
Sorry. I'll clarify, and it can be a different issue (arguably this is a problem with VSCode).


While VSCode supports virtual environments, it does not perform any git actions initiated through its GUI in an activated virtual environment, even if one already exists.

hartwork:
@Spitfire1900 I think it is indeed a separate issue then, thanks for the clarification :+1:

thiagowfx:
You probably meant https://github.com/google/yapf/pull/1093

hartwork:
@thiagowfx indeed, yes, thanks!

Spitfire1900:
Would be part of the idea of this pre-commit --run-all as part of ci.yml?

I would not be opposed to that idea, but I don't know if the output would be very helpful in the event of job failure.

[...]

PR #1093, opened 2023-05-16
Add isort pre-commit and lint with isort and yapf in CI

Spitfire1900:
I repeatedly pushed code as part of working on #1067 that messed up the existing standard isort formatting or failed to follow yapf's own format.

This change

1. Adds a pre-commit to run isort
2. Adds isort configuration that is most-like the existing config `force_single_line=true`, `known_yapftests=yapftests`
3. Has the entire codebase formatted with the above configuration
4. Adds linting steps to CI for isort and yapf
5. Removes hooks from pre-commit  config that was causing the `check-hooks-apply` meta hook to fail

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,3 +32,7 @@ jobs:
         pip install pytest
         pip install pytest-cov
         pytest
+    - name: Lint with isort
+      run: |
+        pip install isort
+        isort . --check --diff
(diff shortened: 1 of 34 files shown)


Spitfire1900:
This PR is ready for review.

thiagowfx (approved):

hartwork:
@Spitfire1900 only saw this now, good stuff, thanks! :+1: 

CC #1081

PR #1096, opened 2023-05-26
Make CI enforce pre-commit clean code

hartwork:
Fixes #1081

Let me know what you think :beers: 

CC @bwendling @Spitfire1900 @thiagowfx

PR #1097, opened 2023-05-26
Make CI keep pre-commit hooks up to date

hartwork:
Was split off of #1096 as suggested at https://github.com/google/yapf/pull/1096#issuecomment-1564727656 .

I'm using a 99% identical approach at multiple other places happily, e.g. pull request https://github.com/hartwork/git-delete-merged-branches/pull/132 was created the same way, also https://github.com/git-big-picture/git-big-picture/pull/318 and so on.

If you like this approach: after the merge, if you trigger workflow `pre_commit_autoupdate` manually it will create a pull request updating the isort pre-commit hook to `v4.4.0` for us. If you don't, it will do that next Friday.

For the CI to be able to create pull requests, this setting on the repository is needed:

permissions

Let me know what you think :beers: 

CC @bwendling @Spitfire1900

### DIFF
--- a/.github/workflows/pre-commit-autoupdate.yml
+++ b/.github/workflows/pre-commit-autoupdate.yml
@@ -1,0 +1,63 @@
+# Copyright (c) 2023 Sebastian Pipping <sebastian@pipping.org>
+# Licensed under the Apache License Version 2.0
+
+name: Keep pre-commit hooks up to date
+
+on:
+  schedule:
+    - cron: '0 16 * * 5'  # Every Friday 4pm
+  workflow_dispatch:
+
+# NOTE: This will drop all permissions from GITHUB_TOKEN except metadata read,
+#       and then (re)add the ones listed below:
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  pre_commit_autoupdate:
+    name: Detect outdated pre-commit hooks
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab  # v3.5.2
+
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-commit-autoupdate.yml
```diff
@@ -0,0 +1,63 @@
+# Copyright (c) 2023 Sebastian Pipping <sebastian@pipping.org>
+# Licensed under the Apache License Version 2.0
+
+name: Keep pre-commit hooks up to date
+
+on:
+  schedule:
+    - cron: '0 16 * * 5'  # Every Friday 4pm
+  workflow_dispatch:
```
Spitfire1900, line 9: Nit-picky, but I don't care for empty assignments, change to an empty dict.
```suggestion
  workflow_dispatch: {}
```

hartwork: I see little value in that personally:
- it works well as is and is not "wrong" in any semantic sense
- and I see no use of dicts in official docs but instead I do see empty assignments, e.g. at https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-of-jobsjob_idsecretsinherit
- it would be inconsistent with what the two existing workflows have

So I consider this addition of noise, personally.

Spitfire1900 (approved):

hartwork:
@Spitfire1900 thanks! :pray:

hartwork:
@bwendling how do you feel about this, any concerns?

bwendling:
This is good. Thank you. I'm sorry for the late reply. I've been on leave this month for family stuff. I made sure the GitHub Actions permissions are what you specified above.

Issue #833, opened 2020-05-01
split_all_comma_separated_values makes force_multiline_dict ineffective


AndydeCleyre:
```console
$ yapf --version
yapf 0.30.0
$ cat .style.yapf
```
```ini
[style]
each_dict_entry_on_separate_line=True
force_multiline_dict=True
split_all_top_level_comma_separated_values=True
```
```console
$ cat test.py
```
```py
COLOR_GROUPS = {
    'Meta-switches': magenta,
    'Switches': yellow,
    'Subcommands': blue
}
```
```console
$ yapf --style .style.yapf -d test.py
```
```diff
--- test.py	(original)
+++ test.py	(reformatted)
@@ -1,5 +1,3 @@
 COLOR_GROUPS = {
-    'Meta-switches': magenta,
-    'Switches': yellow,
-    'Subcommands': blue
+    'Meta-switches': magenta, 'Switches': yellow, 'Subcommands': blue
 }
```

kliyes:
You mean `split_all_comma_separated_values` or `split_all_top_level_comma_separated_values`?

AndydeCleyre:
Whoops, thanks: `split_all_top_level_comma_separated_values`. I'll update the title.

AndydeCleyre:
@brianmego Can you reproduce?

AndydeCleyre:
I've pinpointed the bad behavior at [yapf/yapflib/format_decision_state.py#L201-L204](https://github.com/google/yapf/blob/master/yapf/yapflib/format_decision_state.py#L201-L204), `FormatDecisionState.MustSplit`:

```python
# Allow the fallthrough code to handle the closing bracket.
if current != opening.matching_bracket:
  # If the container doesn't fit in the current line, must split
  return not self._ContainerFitsOnStartLine(opening)
```

Which was formerly just:

```python
# If the container doesn't fit in the current line, must split
return not self._ContainerFitsOnStartLine(opening)
```
Which is problematic in the same way. 5de9ca6e546acb1730b86e982876e9cbe827bc89

brianmego:
Well, I can reproduce this, though I'm not sure what we want to do about it. These two flags are sort of at odds with each other, I think. One says, "Split under all conditions" while the other says "Don't split unless the line is super long". Proof of the issue. This test fails:

https://github.com/brianmego/yapf/commit/52442c1adb7ee2b90471888d2e403bcd47c7143f

AndydeCleyre:
[...]

PR #1103, opened 2023-06-06
SPLIT_ALL_TOP_LEVEL_COMMA_SEPARATED_VALUES and FORCE_MULTILINE_DICT interaction

ManiacDC:
* explicitly defined FORCE_MULTILINE_DICT to override SPLIT_ALL_TOP_LEVEL_COMMA_SEPARATED_VALUES
* added tests to account for the interaction of these two knobs

This is a fix for #833 . This specifically fixes FORCE_MULTILINE_DICT. I think the more correct solution is actually to change this code:
```
      # Allow the fallthrough code to handle the closing bracket.
      if current != opening.matching_bracket:
        # If the container doesn't fit in the current line, must split
        return not self._ContainerFitsOnStartLine(opening)
```

to
```
      # Allow the fallthrough code to handle the closing bracket.
      if current != opening.matching_bracket:
        # If the container doesn't fit in the current line, must split
        if not self._ContainerFitsOnStartLine(opening):
          return True
```

But that is a much more far-reaching change and would change the behavior of existing defaults.


ManiacDC:
CLA fixed, wrong author was on the commit.

dtczest:
Fixed formatting (and CLA, again)

ManiacDC:
Hello, is someone able to run the workflows for this PR? Thanks!

bwendling:
Sorry about the delay. Please add a comment in the CHANGELOG file.

ManiacDC:
@bwendling I've updated the CHANGELOG

ManiacDC:
Thank you @bwendling !

PR #1126, opened 2023-08-25
Feature/logical split

jesse-sony:
First time submitting so there's probably a bunch of stuff I'm doing wrong and we'll have to work out the licensing stuff.

This is a PR to handle the issues in https://github.com/google/yapf/issues/844 where when there is a line that contains logical operators and we split on an operator, we want to split on all operators in that logical line.

I added a new knob to allow this feature to be toggled on. It is off by default for all styles to preserve the existing functioning of the styles. 

I haven't pushed up any tests with this, so that will have to happen before this is merged obviously. However as part of my local testing I ran into some edge cases where you had to be able to set `SPLIT_PENALTY_LOGICAL_OPERATOR` and have it be respected. But the code as it previously existed would ignore penalties if the initial parse set any penalty value even if the desired penalty was higher than what the parse decided on. This prevented the line wrapping from being correct in some extreme edge cases. To get around that, I updated logical line to use the higher `SPLIT_PENALTY_LOGICAL_OPERATOR` but only in the case where the new knob was turned on. I was hesitant to change the normal (existing) behavior of that method, though I do believe that it probably should be considered at some point because higher penalties should probably propagate.


bwendling:
Thanks for working on this! I like the direction you're going here. :-)

jesse-sony:
still working through all the internal hoops to get added to Sony's group that is approved to submit code to Google, but getting closer! Hopefully in next week or so we'll be able to progress on this for real

Issue #1121, opened 2023-07-23
Editable installs on Ubuntu does not add ./third_party dir to PYTHONPATH unless using pip>=23.1


Spitfire1900:
No appropriate `.pth` file is made on Ubuntu using pip<23.1 when installing the project as an editable install.

This means that the following error will occur when running `yapf` as an editable install:
```python
$ yapf --diff --no-local-style matchtest.py 
Traceback (most recent call last):
  File "/home/kyle/Repos/yapf/.wt/1110/.venv/bin/yapf", line 33, in <module>
    sys.exit(load_entry_point('yapf', 'console_scripts', 'yapf')())
  File "/home/kyle/Repos/yapf/.wt/1110/.venv/bin/yapf", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.10/importlib/metadata/__init__.py", line 171, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/kyle/Repos/yapf/.wt/1110/yapf/__init__.py", line 38, in <module>
    from yapf.yapflib import errors
  File "/home/kyle/Repos/yapf/.wt/1110/yapf/yapflib/errors.py", line 16, in <module>
    from yapf_third_party._ylib2to3.pgen2 import tokenize
ModuleNotFoundError: No module named 'yapf_third_party'
```

When installing the project using pip==23.1 the following files are created in `.venv/lib/python3.10/site-packages`

`./.venv/lib/python3.10/site-packages/__editable__.yapf-0.40.1.pth`
```
import __editable___yapf_0_40_1_finder; __editable___yapf_0_40_1_finder.install()
```

`./.venv/lib/python3.10/site-packages/__editable___yapf_0_40_1_finder.py`
[...]

PR #1131, opened 2023-08-31
Migrate to pyproject.toml

Spitfire1900:
Migrate away from setuptools to pyproject.toml

Attanched output of `diff --recursive --unified .venv-setup.py/ .venv-pyproject.toml/ | grep -v 'Binary files' > venv.diff`

[venv.diff.txt](https://github.com/google/yapf/files/12482374/venv.diff.txt)


Spitfire1900:
Testing shows that this closes #1121

PR #1132, opened 2023-08-31
Use python 3.12 for CI 

Spitfire1900:
No particular need for 3.9 - 3.11 (See https://github.com/Spitfire1900/yapf/blob/3.12/yapftests/reformatter_basic_test.py#L25-L26).

Should be ready for 3.12 once it goes GA.


Spitfire1900:
Should we continue to have a different test for each OS<->Python version combination that is warranted or should each OS test all Python versions supported by YAPF?

bwendling:
I think having each OS test all Python versions would be a bit overkill. I'm fine with having one (Linux) test all of them and then the other platforms are tested with a single version (most recent?).

hartwork:
@Spitfire1900 Python 3.12 is available by now.
@bwendling we have checks for >=3.8 and >=3.10 in the code base and you mentioned that Python 3.7 is not dead inside of Google. Taking all that I would vote for covering 3.7, 3.8, 3.10, 3.12 but not 3.9 and not 3.11 to save resources.

Spitfire1900:
>=3.10 means that 3.9-3.11 is not applicable, and really only testing on 3.7, 3.8, and 3.12 is required.

hartwork:
> `>=3.10` means that 3.9-3.11 is not applicable, and really only testing on 3.7, 3.8, and 3.12 is required.

@Spitfire1900 I don't understand. Could you elaborate? On my part: if we check for >=3.10 but test with say 3.12 but not 3.10, technically we have no guarantee that that the code actually works with 3.10 and doesn't e.g. require >=3.11 in reality.

Spitfire1900:
Fair. Technically verify `>3.10` but not equality.

PR #1134, opened 2023-09-04
pyproject.toml updates to HACKING.md

Spitfire1900:
* Setup module is missing, use build instead.
* Run tests with tox.


Spitfire1900:
@bwendling after this is merged in I think we're in a good spot to release 0.41.0

bwendling:
Could you add example commands for each step?

bwendling:
Yeah, something like that would be fine.

Spitfire1900:
@bwendling I've updated `HACKING.md` and I've confirmed that this works on both Ubuntu 22.04 and Windows 10.

bwendling (approved):

PR #1150, opened 2023-09-23
copy globals()

### LINKED ISSUE #1118: error is raised at `for _name, _value in list(globals().items()):` in yapf_third_party/_ylib2to3/pgen2/token.py
Hi, I got an error when importing yapf whose version is 0.40.1
When the yapf is imported, yapf_third_party/_ylib2to3/pgen2/token.py is executed
and then error is raised as below.
image
I think it occurs to other people also https://github.com/scipy/scipy/issues/11479
Something weird is it appears not in all cases.
It can be solved easily if `for _name, _value in list(globals().items()):` is changed to `for _name, _value in list(globals().copy().items()):`.
Could you fix this bug?

Spitfire1900:
Fixes #1118. Use a copy of globals() rather than a direct reference. This did not appear to impact the time of tests.

### DIFF
--- a/third_party/yapf_third_party/_ylib2to3/pgen2/token.py
+++ b/third_party/yapf_third_party/_ylib2to3/pgen2/token.py
@@ -70,7 +70,7 @@ NT_OFFSET = 256
 # --end constants--
 
 tok_name = {}
-for _name, _value in list(globals().items()):
+for _name, _value in list(globals().copy().items()):
   if isinstance(_value, int):
     tok_name[_value] = _name
 
--- a/third_party/yapf_third_party/_ylib2to3/pgen2/token.py
+++ b/third_party/yapf_third_party/_ylib2to3/pgen2/token.py
@@ -69,10 +69,11 @@ N_TOKENS = 60
 NT_OFFSET = 256
 # --end constants--
 
-tok_name = {}
-for _name, _value in list(globals().copy().items()):
-  if isinstance(_value, int):
-    tok_name[_value] = _name
+tok_name = {
+    _value: _name
+    for _name, _value in globals().copy().items()
+    if isinstance(_value, int)
+}
 
 
 def ISTERMINAL(x):


Spitfire1900:
Requesting review from @bwendling and @char101 .
Requesting testing from @eunwoosh.

Spitfire1900:
@bwendling , it doesn't look like I can request reviewers within the GitHub web UI like is documented here. Is there some project setting that needs to be updated?


[eunwoosh](https://github.com/eunwoosh)

Spitfire1900:
> @bwendling , it doesn't look like I can request reviewers within the GitHub web UI like is documented here. Is there some project setting that needs to be updated?
> 
> image

A user needs to be marked as a collaborater to request a reviewer through the WEB UI.

char101:
If you copy the `globals()`, isn't the `list()` part not required anymore?

char101:
Probably the module is imported twice at the same time, the first import finished a bit faster, and then it adds `ISTERMINAL`, etc. to the global scopy, which broke the second import which is still looping. We could also fix this by moving the loop to the bottom of the file.

Spitfire1900:
So we're using 3.7, so I've elected to use copy() and dictionary comprehension here, while dropping the list() wrapper.

Spitfire1900:
> Probably the module is imported twice at the same time, the first import finished the loop a bit faster, and then it adds `ISTERMINAL`, etc. to the global scope, which broke the second import which is still looping. We could also fix this by moving the loop to the bottom of the file.

This makes logical sense... a bit. I may be out of the loop here but I would think that if a module is imported twice the first import blocks until initialization completes.

eunwoosh:
> Requesting review from @bwendling and @char101 . Requesting testing from @eunwoosh.

I tested and the error isn't raised anymore. Thanks for fixing it.

bwendling (approved):

PR #1152, opened 2023-09-24
Use tomllib on Python 3.11+

felixonmars:
Fallback to tomli for compatibility with older Python

### DIFF
--- a/yapf/yapflib/file_resources.py
+++ b/yapf/yapflib/file_resources.py
@@ -52,11 +52,14 @@ def _GetExcludePatternsFromPyprojectToml(filename):
   """Get a list of file patterns to ignore from pyproject.toml."""
   ignore_patterns = []
   try:
-    import tomli as tomllib
+    import tomllib
   except ImportError:
-    raise errors.YapfError(
-        'tomli package is needed for using pyproject.toml as a '
-        'configuration file')
+    try:
+      import tomli as tomllib
+    except ImportError:
+      raise errors.YapfError(
+          'tomli package is needed for using pyproject.toml as a '
+          'configuration file')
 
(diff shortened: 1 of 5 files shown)


Spitfire1900:
Would you like to do similar for `importlib-metadata` as part of this PR? The PyPi package is only required on Python 3.7.

yapf/yapflib/file_resources.py
```diff
@@ -52,11 +52,14 @@ def _GetExcludePatternsFromPyprojectToml(filename):
   """Get a list of file patterns to ignore from pyproject.toml."""
   ignore_patterns = []
   try:
-    import tomli as tomllib
```
Spitfire1900, line 55: Not a fan of nested try/catches in favor of https://peps.python.org/pep-0484/#version-and-platform-checking style approach.

I'm also wondering if moving the import to ./yapf/yapflib/__init__.py is a good idea idea vs. conditional login in every file.

Lastly, this package is now part of the package requirements, so managing an `ImportError` may no longer be applicable.

```python
import sys
if sys.version_info >= (3, 10):
    import tomllib
else:
    import tomli as tomllib
```

if ImportError must still be handled.
```python
import sys
if sys.version_info >= (3, 10):
    import tomllib
else:
    try:
        import tomli as tomllib
    except ImportError:
        raise errors.YapfError(
            'tomli package is needed for using pyproject.toml as a '
            'configuration file')
```

Spitfire1900 (changes requested):
Use sys.version_info instead of nested try/catches.

Strip all import try/catches as tomli / tomllib is guaranteed by requirements specification.

felixonmars:
Looks like requested changes are in #1158 now. Closing in favor of that one.

PR #1177, opened 2023-11-02
Add new DISABLE_SPLIT_LIST_WITH_COMMENT flag.

jlebar:
`DISABLE_SPLIT_LIST_WITH_COMMENT` is a new knob that changes the
behavior of splitting a list when a comment is present inside the list.

Before, we split a list containing a comment just like we split a list
containing a trailing comma: Each element goes on its own line (unless
`DISABLE_ENDING_COMMA_HEURISTIC` is true).

Now, if `DISABLE_SPLIT_LIST_WITH_COMMENT` is true, we do not split every
element of the list onto a new line just because there's a comment somewhere
in the list.

This mirrors the behavior of clang-format, and is useful for e.g. forming
"logical groups" of elements in a list.

Note: Upgrading will result in a behavioral change if you have
`DISABLE_ENDING_COMMA_HEURISTIC` in your config.  Before this version, this
flag caused us not to split lists with a trailing comma *and* lists that
contain comments.  Now, if you set only that flag, we *will* split lists
that contain comments.  Set the new `DISABLE_SPLIT_LIST_WITH_COMMENT` flag to
true to preserve the old behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - Remove dependency on importlib-metadata
 - Remove dependency on tomli when using >= py311
 ### Added
+
+#### New `DISABLE_SPLIT_LIST_WITH_COMMENT` flag
 - `DISABLE_SPLIT_LIST_WITH_COMMENT` is a new knob that changes the
   behavior of splitting a list when a comment is present inside the list.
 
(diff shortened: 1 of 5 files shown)


jlebar:
cc @ptillet

CHANGELOG.md
```diff
@@ -6,6 +6,28 @@
...
+  in the list.
+
+  This mirrors the behavior of clang-format, and is useful for e.g. forming
+  "logical groups" of elements in a list.
+
+  Note: Upgrading will result in a behavioral change if you have
+  `DISABLE_ENDING_COMMA_HEURISTIC` in your config.  Before this version, this
+  flag caused us not to split lists with a trailing comma *and* lists that
+  contain comments.  Now, if you set only that flag, we *will* split lists
+  that contain comments.  Set the new `DISABLE_SPLIT_LIST_WITH_COMMENT` flag to
```
bwendling, line 28: This confuses me. You say that with this flag:

> we will split lists that contain comments

But isn't that exactly what this flag is supposed to prevent?

jlebar: I think it's right?

> You say that with this flag:
>
>> we will split lists that contain comments

That's not what I intended to convey.  In the context above, "this flag" refers to `DISABLE_ENDING_COMMA_HEURISTIC`.

To be more clear, here's what changes.

- Before:
    - Default: split lists with a trailing comma or with comments. 
    - `DISABLE_ENDING_COMMA_HEURISTIC=true`, do not split lists with a trailing comma or with comments.

- After:
    - Default: Same as before.
    - `DISABLE_ENDING_COMMA_HEURISTIC=true` and `DISABLE_SPLIT_LIST_WITH_COMMENT=false`, do not split lists with trailing comma, but do split lists with comments.  *This is different than before.*
    - `DISABLE_ENDING_COMMA_HEURISTIC=false` and `DISABLE_SPLIT_LIST_WITH_COMMENT=true`, do split lists with a trailing comma, but do not split lists with comments.
    - Both flags true: do not split lists with a trailing comma, and do not split lists with comments.

(I hope I got that right above.)

jlebar:
Thank you for the review!

CHANGELOG.md
bwendling: Ah! Okay.

Could you add the interaction with `DISABLE_ENDING_COMMA_HEURISTIC` in the README.md and style.py descriptions? I want to make sure people know about any any unforeseen side-effects. :-)

jlebar: Done, please have a look.  And thanks again!

bwendling (approved):
Thank you for this excellent work. And also for adding documentation for `DISABLE_SPLIT_LIST_WITH_COMMENT`. :-)

jlebar:
Woo, thank you for merging!

jlebar:
Note to self: This branch is referred to in the Triton pre-commit, so I should not delete it.  :)

Issue #1170, opened 2023-10-12
 Support 3.12 feature: PEP 695 Type Parameter Syntax


joshlam123:
### Issue
In [PEP-695](https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep695), running under 3.12 will cause exceptions in YAPF. For example the following two lines of code for testing:

```
type Testing = dict[int | str]
type Testing2 = dict[str: int]
```

<b> The following error message happens:</b>

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/josh.lam/.local/yapf/yapf/__main__.py", line 18, in <module>
    yapf.run_main()
  File "/home/josh.lam/local/.pyenv/versions/3.12.0/lib/python3.12/site-packages/yapf/__init__.py", line 377, in run_main
    sys.exit(main(sys.argv))
             ^^^^^^^^^^^^^^
  File "/home/josh.lam/.local/.pyenv/versions/3.12.0/lib/python3.12/site-packages/yapf/__init__.py", line 136, in main
    changed = FormatFiles(
              ^^^^^^^^^^^^
  File "/home/josh.lam/.local/.pyenv/versions/3.12.0/lib/python3.12/site-packages/yapf/__init__.py", line 214, in FormatFiles
    changed |= _FormatFile(filename, lines, style_config, no_local_style,
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/josh.lam/local/.pyenv/versions/3.12.0/lib/python3.12/site-packages/yapf/__init__.py", line 248, in _FormatFile
    raise errors.YapfError(errors.FormatErrorMsg(e))
                           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/josh.lam/.local/.pyenv/versions/3.12.0/lib/python3.12/site-packages/yapf/yapflib/errors.py", line 37, in FormatErrorMsg
    return '{}:{}:{}: {}'.format(e.args[1][0], e.args[1][1], e.args[1][2], e.msg)
                                 ~~~~~~^^^
IndexError: tuple index out of range
>>>>> In file /home/josh.lam/testing.py <<<<<
1,2d0
< type Testing = dict[int | str]
< type Testing2 = dict[str: int]
```

[...]

PR #1179, opened 2023-11-08
Remark PEP 695 as an unsupported feature

Spitfire1900:
For issue #1170.

Remark remark PEP 695 as an unsupported feature


bwendling (approved):

wookayin:
Is this "Unsupported" or "yet-to-be-supported"?

PR #1254, opened 2024-12-10
(fix) split_arguments_when_comma_terminated: skip comment

alexey-pelykh:
Fixes issue that the following code won't be reformatted with `split_arguments_when_comma_terminated` enabled:

```python
class Class(  # comment
        object,):
    pass
```


alexey-pelykh:
(I have not added the CHANGELOG as the file seems outdated given the currently released version)

PR #1243, opened 2024-10-03
Fix pickle related race condition (fixes #1164, fixes #1204)

hartwork:
Fixes #1164
Fixes #1204

I have flaky pre-commit CI caused by Yapf in multiple repositories by now and it was about time to start doing something about it… I'm grateful for the great analysis by both @a-gardner1 and @whlook.

~The pull request sits on top of #1242 for the moment purely for a chance at a green CI: I'm happy to rebase off of it after that one has been merged.~

For a convenient reproducer of the race condition based on earlier work by @whlook please see https://github.com/google/yapf/issues/1204#issuecomment-2391417931 .

If this could be reviewed and merged soon-ish, that would rock the house. Thanks in advance! :pray: 

CC @bwendling @whlook @a-gardner1 @kamahen

PR #1263, opened 2025-02-05
Add an explicit `flush` and `close` before `os.rename` to fix `EOFError: Ran out of input`

anmyachev:
Even with this fix: https://github.com/google/yapf/pull/1243 `EOFError: Ran out of input` error is observed when using `yapf` with `pre-commit` on powerful multi-core systems.

Before I suggest this change to you, I tested it in our repository using a fork: https://github.com/intel/intel-xpu-backend-for-triton/pull/3324. For 5 days now I have not seen the error again, which gives me confidence to say that the change at least reduces the frequency of this problem, and at most solves it completely.

Hi @bwendling, could you review this change?

### DIFF
--- a/third_party/yapf_third_party/_ylib2to3/pgen2/grammar.py
+++ b/third_party/yapf_third_party/_ylib2to3/pgen2/grammar.py
@@ -111,6 +111,8 @@ class Grammar(object):
         dir=tempfile_dir,
         delete=False) as f:
       pickle.dump(self.__dict__, f.file, pickle.HIGHEST_PROTOCOL)
+      f.flush()
+      f.close()
       try:
         os.rename(f.name, filename)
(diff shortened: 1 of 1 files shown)


anmyachev:
Hi @bwendling!

`EOFError: Ran out of input` error has not been observed for 2 weeks now (with this fix). Friendly reminder to review.

Hi @hartwork, maybe you can review too.

alexkau:
My team recently had to upgrade yapf to get python 3.13 compatibility, and we're now seeing this issue once or twice per week. We aren't using especially high-end build nodes (16 cpu, 32GB RAM), and we're not an especially large team. We're going to try switching to the forked version with this fix, but it seems like this issue might be widespread.

Issue #1258, opened 2025-01-05
Replatform off of lib2to3 to libcst and structured pattern matching (2025 project idea)


Spitfire1900:
YAPF migrated to a fork of lib2to3 as _ylib2to3 in order to patch in support for 3.10 features like structured pattern matching. 

Maintaining the fork of lib2to3 as a long term project to implement 3.12+ new language features insn't feasible and should consider a separate concrete syntax tree (CST) provider like libcst.

This year also marks the EoL of Python 3.9, so structured pattern matching will be available in all supported versions of Python come November; this re-platforming effort may take advantage of that also.

CTRLRLTY:
Is this currently being worked on?

CTRLRLTY:
I'll assume no one is on it. Im gonna implement this idea on a new fork.

Spitfire1900:
@CTRLRLTY It is not yet being worked on. I do not have the resources to work on it.

PR #1265, opened 2025-02-07
yapf will fail on Python >= 3.13 because lib2to3 was removed

cclauss:
`lib2to3` has been removed from the standard library in Python 3.13.

Document the incompatibility and the need for:
* #1258

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.11", "3.12"]  # no particular need for 3.9 or 3.10
+        # Python >= 3.13 will fail with `ModuleNotFoundError: No module named 'lib2to3'`
+        # See: google/yapf#1258
+        python-version: ["3.8", "3.11", "3.12", "3.13"]  # no particular need for 3.9 or 3.10
         os: [macos-latest, ubuntu-latest, windows-latest]
     steps:
     - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871  # v4.2.1


Spitfire1900:
This isn't true, lib2to3 is no longer used by YAPF since #1067 .

The project does still need to move off even the forked version though because it's not feasible to patch [_ylib2to3](https://github.com/google/yapf/tree/7e218233d634829795987465d6cd6ba9779c6514/third_party/yapf_third_party/_ylib2to3) with new Python language features.


```diff
@ python --version; python -m yapf --diff get-pip.py
Python 3.14.0a1
--- get-pip.py  (original)
+++ get-pip.py  (reformatted)
@@ -28,11 +28,11 @@
     message_parts = [
         "This script does not work on Python {}.{}.".format(*this_python),
         "The minimum supported Python version is {}.{}.".format(*min_version),
-        "Please use https://bootstrap.pypa.io/pip/{}.{}/get-pip.py instead.".format(*this_python),
+        "Please use https://bootstrap.pypa.io/pip/{}.{}/get-pip.py instead.".
+        format(*this_python),
     ]
     print("ERROR: " + " ".join(message_parts))
     sys.exit(1)
-

 import os.path
 import pkgutil
@@ -28835,6 +28835,5 @@
 a4%nWWo~3|axZmqY;0*_GcR>?X>2cdVQF+OaCuNm1qJ{B006H6uL0$R000pt8vp<R
 """

-
 if __name__ == "__main__":
     main()
```

cclauss:
Awesome.  Thanks much for your work on this.

.github/workflows/ci.yml
```diff
@@ -13,7 +13,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.11", "3.12"]  # no particular need for 3.9 or 3.10
+        # Python >= 3.13 will fail with `ModuleNotFoundError: No module named 'lib2to3'`
+        # See: google/yapf#1258
+        python-version: ["3.8", "3.11", "3.12", "3.13"]  # no particular need for 3.9 or 3.10
```
cclauss, line 18: It still might be useful to add Py313 to the testing.

Spitfire1900: Can probably drop 3.11 and 3.12 from the list. There's some special cases with 3.8 left in implementation code but 3.13 functionally tests everything >3.10 or greater.

There's also no special 3.9 code.

stav-bitanski:
@cclauss @Spitfire1900 it crashes for me too, what is the fix?

Spitfire1900:
@stav-bitanski see https://github.com/google/yapf?tab=readme-ov-file#python-features-not-yet-supported , you may be using a feature of Python 3.12+ that is not yet supported by YAPF.

PR #1297, opened 2025-12-23
Improve support for ignoring files.

matts1:
It is now possible to run yapf on ignored files without it erroring out, and to pass ignored files to stdin.


matts1:
As an example, in depot_tools ([source](https://source.chromium.org/chromium/chromium/src/+/main:third_party/depot_tools/git_cl.py;l=746-773;drc=04f3706a63d580fb0a9affef9d6c020c8bc7ac3c)), we actually parse the .yapfignore files ourselves because otherwise it errors out when we pass it ignored files.

PR #1320, opened 2026-09-10
fix(reformatter): ensure idempotent formatting for dictionary values with line continuation

### LINKED ISSUE #1217: yapf adds spaces before backslash with every run
Let's say I have this code fragment (note the backslash):

```python3
plan_bundles = {
    plan_bundle.bundle_name: {
        'tier_change_options': \
        bundle_util.get_tier_change_options(plan_bundle, details_by_bundle)
    } for plan_bundle in plan.plan_bundles
}
```
If I run yapf over that file it will add two spaces before the backlash. Result:
image

If I run it again:
image

So yapf is never happy with my file code style. Is this a bug?

Style config:
```
[style]
based_on_style = google
indent_width = 2
column_limit = 100
continuation_indent_width = 4
each_dict_entry_on_separate_line = true
allow_split_before_dict_value = false
split_before_logical_operator = true
spaces_before_comment = 1
blank_line_before_nested_class_or_def = false
```

jdymitarai:
### Summary
Fixes #1217 where yapf repeatedly added spaces before a line continuation backslash (`\`) in dictionary values on subsequent formatting runs (under styles enabling `INDENT_DICTIONARY_VALUE`, such as `google`). Also resolves an `AssertionError` when dictionary values consisting of a single leaf were preceded by a line continuation backslash.

### Cause & Analysis
1. **Continuation node parentage**: In `continuation_splicer.py`, continuation nodes were inserted into pytree nodes via `node.children.insert()` directly instead of `node.insert_child()`, leaving `continuation_node.parent` as `None`. Any subsequent AST transformations or sibling queries (`next_sibling`, `replace`) on these nodes could fail or raise an `AssertionError`.
2. **Continuation in dictionary values**: In `subtype_assigner.py`, `_InsertPseudoParentheses` only checked `grammar_token.COMMENT` when looking for tokens inserted before a value leaf, missing `format_token.CONTINUATION`. As a result, single-leaf values preceded by a backslash failed to be encompassed into an atom node alongside the continuation marker.
3. **Redundant whitespace emission**: In `reformatter.py`, `_FormatFinalLines` appends a space for pseudo-parentheses if the succeeding token's `whitespace_prefix` does not start with whitespace. Because continuation tokens hold their leading whitespace in their token value (`' \'`) rather than `whitespace_prefix`, `_FormatFinalLines` emitted an additional space before the continuation token on every formatting run, causing infinite drift.
4. **Token space calculation**: In `logical_line.py`, continuation tokens were checked after pseudo-parentheses in `_SpaceRequiredBetween`, allowing pseudo scope opening to request an unnecessary space before continuation markers.

### Solution
- Use `node.insert_child(...)` in `continuation_splicer.py` to maintain consistent parent links.
- Handle `format_token.CONTINUATION` alongside `grammar_token.COMMENT` in `_InsertPseudoParentheses`.
- Check `not tok.next_token.value.startswith(' ')` in `_FormatFinalLines` before appending spaces for pseudo tokens.
- Prioritize continuation token spacing checks in `logical_line._SpaceRequiredBetween`.
- Add unit tests in `reformatter_basic_test.py` verifying multiple-pass idempotency and single-leaf dictionary value continuation handling.

### DIFF
--- a/yapf/pytree/continuation_splicer.py
+++ b/yapf/pytree/continuation_splicer.py
@@ -46,7 +46,7 @@ def SpliceContinuations(tree):
     for index, child in enumerate(node.children[:]):
       continuation_node = RecSplicer(child)
       if continuation_node:
-        node.children.insert(index + num_inserted, continuation_node)
+        node.insert_child(index + num_inserted, continuation_node)
         num_inserted += 1
 
   RecSplicer(tree)
--- a/yapf/pytree/subtype_assigner.py
+++ b/yapf/pytree/subtype_assigner.py
@@ -30,6 +30,7 @@ from yapf_third_party._ylib2to3.pygram import python_symbols as syms
 
 from yapf.pytree import pytree_utils
 from yapf.pytree import pytree_visitor
+from yapf.yapflib import format_token
 from yapf.yapflib import style
 from yapf.yapflib import subtypes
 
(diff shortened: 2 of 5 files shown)


jdymitarai:
Closing this PR to avoid review overhead on the team. The context remains in the thread for future reference if helpful. Thanks for your time.

Issue #1256, opened 2024-12-25
IndexError on type definition


hoijui:
yapf version: `0.43.0`
python3.12

to reproduce:

```shell
$ echo "type foo = str" | yapf
Traceback (most recent call last):
  File "/home/user/.cache/pypoetry/virtualenvs/krawl-_WstNwwR-py3.12/lib/python3.12/site-packages/yapf/yapflib/yapf_api.py", line 198, in FormatCode
    tree = pytree_utils.ParseCodeToTree(unformatted_source)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/krawl-_WstNwwR-py3.12/lib/python3.12/site-packages/yapf/pytree/pytree_utils.py", line 113, in ParseCodeToTree
    tree = parser_driver.parse_string(code, debug=False)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/krawl-_WstNwwR-py3.12/lib/python3.12/site-packages/yapf_third_party/_ylib2to3/pgen2/driver.py", line 187, in parse_string
    return self.parse_tokens(tokens, debug)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/krawl-_WstNwwR-py3.12/lib/python3.12/site-packages/yapf_third_party/_ylib2to3/pgen2/driver.py", line 156, in parse_tokens
    if p.addtoken(type, value, (prefix, start)):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/krawl-_WstNwwR-py3.12/lib/python3.12/site-packages/yapf_third_party/_ylib2to3/pgen2/parse.py", line 230, in addtoken
    return self._addtoken(ilabel, type, value, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/krawl-_WstNwwR-py3.12/lib/python3.12/site-packages/yapf_third_party/_ylib2to3/pgen2/parse.py", line 313, in _addtoken
[...]

Issue #1250, opened 2024-10-17
Crash on format strings with un-escaped quotes in the format element


goldsteinn:
Take `test.py`:
```
import os
print(f"foo -> {os.path.join("bar", "baz")}")
```

`python3 test.py` runs successfully and produces `foo -> bar/baz`

`yapf test.py` crashes with:
```
Traceback (most recent call last):
  File "yapf/yapf/yapflib/yapf_api.py", line 198, in FormatCode
    tree = pytree_utils.ParseCodeToTree(unformatted_source)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "yapf/yapf/pytree/pytree_utils.py", line 113, in ParseCodeToTree
    tree = parser_driver.parse_string(code, debug=False)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".local/lib/python3.12/site-packages/yapf_third_party/_ylib2to3/pgen2/driver.py", line 187, in parse_string
    return self.parse_tokens(tokens, debug)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".local/lib/python3.12/site-packages/yapf_third_party/_ylib2to3/pgen2/driver.py", line 156, in parse_tokens
    if p.addtoken(type, value, (prefix, start)):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".local/lib/python3.12/site-packages/yapf_third_party/_ylib2to3/pgen2/parse.py", line 230, in addtoken
    return self._addtoken(ilabel, type, value, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".local/lib/python3.12/site-packages/yapf_third_party/_ylib2to3/pgen2/parse.py", line 313, in _addtoken
    raise ParseError('bad input', type, value, context)
yapf_third_party._ylib2to3.pgen2.parse.ParseError: bad input: type=1, value='bar', context=('', (2, 32))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "yapf/yapf/__init__.py", line 238, in _FormatFile
    reformatted_code, encoding, has_change = yapf_api.FormatFile(
                                             ^^^^^^^^^^^^^^^^^^^^
[...]

Issue #1215, opened 2024-03-20
Ternary conditionals within F Strings Raise IndexError


CS-Account:
**Example:**

yapf_example.py contents:
```python yapf_example.py

print(f"This Fails, but is legal: {"Should be fine" if True else "But it's not"}")
```

**Running Yapf:**
`python -m yapf yapf_example.py`

**Result:**

```
...
    return '{}:{}:{}: {}'.format(e.args[1][0], e.args[1][1], e.args[1][2], e.msg)
                                 ~~~~~~^^^
IndexError: tuple index out of range
```

**Yapf version**
0.40.2 (Latest at time of writing)

This might be related to [PEP 701](https://github.com/google/yapf/issues/1136), but if memory serves, this has been an issue since before 3.12 as well.

PR #1321, opened 2026-09-10
fix(errors): handle parse.ParseError without IndexError in FormatErrorMsg

### LINKED ISSUE #1303: [Bug][Crash][Reproducible] IndexError: tuple index out of range
# Problems
YAPF fails with `IndexError: tuple index out of range` while parsing a simple file
# Reproduction
 - `yapf==0.43.0`
 - Put this in `test.py`: `f"{tab["SOME_STRING"]}"`
 - run `yapf test.py`

jdymitarai:
### Summary
Fixes #1303, #1250, #1215, #1256 where YAPF crashed with an unhandled internal `IndexError: tuple index out of range` whenever parsing raised a `parse.ParseError` instead of producing a clean error diagnostic.

### Cause & Analysis
In `yapf/pytree/pytree_utils.py`, when parsing fails, `ast.parse(code)` is executed to check if the snippet is invalid Python syntax (raising `SyntaxError`). If `ast.parse` succeeds (for example on newer language constructs or syntax supported by Python's parser but not by the `lib2to3` grammar), `ParseCodeToTree` re-raises `parse.ParseError`.

When handling this exception in `yapf_api.py`, `FormatErrorMsg(e)` attempted to format the error location using `e.args[1][0], e.args[1][1], e.args[1][2]`, assuming `e.args[1]` was a 3-element tuple `(filename, lineno, column)`. However, `_ylib2to3.pgen2.parse.ParseError.args` is a 1-tuple `("bad input: ...",)` where location metadata is stored in `e.context` (`(prefix, (lineno, column))`). This resulted in `e.args[1]` raising an unhandled `IndexError: tuple index out of range` instead of reporting the syntax error cleanly.

### Solution
1. Explicitly recognize `parse.ParseError` in `FormatErrorMsg`:
   - Extract `lineno` and `column` from `e.context[1]` if available.
   - Extract the error message from `e.msg`.
2. Consistently use `getattr(e, 'filename', None)` across formatters.
3. Add a safe fallback block in `FormatErrorMsg` to handle unexpected exception argument shapes cleanly without raising `IndexError`, `TypeError`, or `AttributeError`.
4. Add unit tests for `ParseError` formatting and `FormatErrorMsg` in `yapftests/yapf_test.py`.

### DIFF
--- a/yapf/yapflib/errors.py
+++ b/yapf/yapflib/errors.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """YAPF error objects."""
 
+from yapf_third_party._ylib2to3.pgen2 import parse
 from yapf_third_party._ylib2to3.pgen2 import tokenize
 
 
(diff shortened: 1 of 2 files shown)


jdymitarai:
Closing this PR to avoid review overhead on the team. The context remains in the thread for future reference if helpful. Thanks for your time.

